PR #10, opened 2021-02-01
Configure networking with os-net-config and create 2 provider networks

mdbooth:


playbooks/prepare_stack_testconfig.yaml
```diff
@@ -0,0 +1,92 @@
...
+          openstack security group create basic
+      fi
+      if ! openstack security group rule list basic | grep "22:22"; then
+          openstack security group rule create basic --protocol tcp --dst-port 22:22 --remote-ip 0.0.0.0/0
+      fi
+      if ! openstack security group rule list basic | grep "icmp"; then
+          openstack security group rule create --protocol icmp basic
+      fi
+      if ! openstack security group rule list basic | grep "53:53"; then
+          openstack security group rule create --protocol udp --dst-port 53:53 basic
```
EmilienM, line 55: since it's OUTPUT rule, I guess it's already open by default, isn't? I never had to create such a rule, I wonder why you added it here.

playbooks/prepare_stack_testconfig.yaml
```diff
@@ -0,0 +1,92 @@
...
+  - name: Copy ssh public key to host
+    copy:
+      src: "{{ testconfig_public_key }}"
+      dest: "{{ ansible_env.HOME }}/.ssh/{{ testconfig_public_key | basename }}"
+      owner: stack
+      group: stack
+      mode: '0600'
+    register: testconfig_public_key_copy
+
+  - name: Configure resources for openshift user
```
EmilienM, line 31: ideally we would use ```OpenStack Cloud``` collections, e.g. [keypair](https://docs.ansible.com/ansible/latest/collections/openstack/cloud/keypair_module.html#ansible-collections-openstack-cloud-keypair-module), etc. but it's just an FYI.

playbooks/roles/network_info/tasks/main.yaml
```diff
@@ -0,0 +1,30 @@
+---
+- name: Read network info if present
+  slurp:
+    src: &networkinfo /etc/dev-install-network-info.yaml
```
EmilienM, line 4: why do you use anchors?

playbooks/prepare_stack_testconfig.yaml
mdbooth: I just copied this 😂 I have no idea why it's there tbh.

playbooks/prepare_stack_testconfig.yaml
mdbooth: Agreed. I've had a poor experience with those in the past, though, a combination of being both buggy and incomplete. On a previous project I started using them exclusively, then switched to cli for the stuff which didn't work or wasn't possible, then decided the complication of supporting 2 methods wasn't worth it and switched to cli for everything.

Ideally I'd go all in and just fix/improve the ansible. I never seem to be working on something where that seems worthwhile, although I'm aware that's short term thinking.

playbooks/roles/network_info/tasks/main.yaml
mdbooth: Only to DRY where it's not worth, for eg, setting an ansible fact. I don't use them very consistently, though.

mandre:
Looks good. Do you want to merge it now or give it more time?

mdbooth:
> Looks good. Do you want to merge it now or give it more time?

I know @adduarte has concerns about the external nic being on any bridge, even one that isn't configured for use by neutron. I was hoping to get him on board before merging. I'm not personally convinced there's an issue here and I prefer the simplicity of just having a single network configuration. However, I also want @adduarte to be able to use it. If the complexity is genuinely required I'll have to look at it next time I'm hacking on this.

playbooks/network.yaml
```diff
@@ -99,41 +61,23 @@
...
-        state: started
-        daemon_reload: yes
-      loop:
-      - standalone_public_route
-      - standalone_public_snat
-    when: not public_uses_external_nic
+  - name: Activate systemd units
+    systemd:
+      name: "{{ item }}.service"
+      enabled: yes
```
EmilienM, line 79: yes is deprecated everywhere in ansible, FYI, we use booleans now, here "true"

playbooks/prepare_stack_testconfig.yaml
```diff
@@ -0,0 +1,92 @@
...
+          openstack keypair create --public-key {{ testconfig_public_key_copy.path }} default
+      fi
+      if ! openstack network show private >/dev/null; then
+          openstack network create --internal private
+      fi
+      if ! openstack subnet show private-subnet >/dev/null; then
+          openstack subnet create private-subnet --subnet-range {{ testconfig_private_cidr }} --network private
+      fi
+      # create basic security group to allow ssh/ping/dns
+      if ! openstack security group show basic >/dev/null; then
```
adduarte, line 45: In all of this openstack show calls you can probably just execute with a flag for the show command. 
I guess it makes no difference, but it lowers need traffic back and forth on the client

adduarte: Also I see a few calls to openstack that the ansible cloud module would exectue as well. 
Maybe I am missing the context of where this is executed. ( 
For example this whole shell command could be broken down into ansible calls, unless it is being executed on a remote machine (removed twice from ansible-playbook is being executed).

playbooks/prepare_stack_testconfig.yaml
adduarte: I am not sure if i am following the context of where this rules are being created but if Openstck security groups are usually created with source/dest requirements.  And they might not be necessary OUTPUT. 
it could be that thi is a security group to be used bewteen hosts in the same subnet (tenant). 
we do the same in openshift setup because the dns server (port 53) and the workers are in the same subnet, but they are not allowed to talk to each other unless a rule exists in the security group that controls their traffic

playbooks/prepare_stack_testconfig.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+  - name: Configure resources for openshift user
+    shell: |
+      set -e -o pipefail
+
+      if ! openstack keypair show default >/dev/null; then
+          openstack keypair create --public-key {{ testconfig_public_key_copy.path }} default
+      fi
+      if ! openstack network show private >/dev/null; then
+          openstack network create --internal private
```
adduarte, line 39: you can use a flag on the openstck commands to modified their output to lower traffic back and forth. some cli commands result in several rest calls to the openstack api , for example you coud add something like   -f value -c id, 
and theonly thing you should get is the uuid of the object that was created. 

I guess it not much, but i just hate that dam pause when executing an openstack cli.

playbooks/prepare_stack.yaml
```diff
@@ -27,16 +30,31 @@
...
+              --network external
+      fi
+    environment:
+      OS_CLOUD: standalone
+    when: external_fip_pool_start is defined and external_fip_pool_end is defined
+
+  - name: Create hostonly network
+    shell: |
+      if ! openstack network show hostonly; then
+          openstack network create --external --share --provider-physical-network hostonly --provider-network-type flat hostonly
```
EmilienM, line 51: --project openshift

playbooks/prepare_stack.yaml
```diff
@@ -27,16 +30,31 @@
...
+      fi
+      if ! openstack subnet show hostonly-subnet; then
+          openstack subnet create hostonly-subnet --subnet-range "{{ hostonly_cidr }}" \
```
EmilienM, line 54: --project openshift

playbooks/network.yaml
```diff
@@ -99,41 +61,23 @@
...
-        group: root
-        mode: '0644'
-      vars:
-        cidr: "{{ public_cidr }}"
-        exit_if: "{{ ansible_default_ipv4.interface }}"
-        exit_ip: "{{ ansible_default_ipv4.address }}"
+  - name: Create systemd unit to add SNAT rule for hostonly network
+    template:
+      src: "standalone_snat.service.j2"
+      dest: "/etc/systemd/system/standalone_hostonly_snat.service"
```
EmilienM, line 67: so in that file, I had to replace {{ exit_if }} by br-ex and it worked fine.

mdbooth: Thanks! I can even hard-code this because we always create br-ex with the external interface bridged, so this is a super simple fix.

playbooks/network.yaml
mdbooth: TIL!

That will take as long as python2 deprecation to clean up at this point, though 😂

playbooks/prepare_stack.yaml
mdbooth: Do we want to parameterise this or not? dev-install is unapologetically for use by our team, so if you tell me we always want this I'm fine hardcoding it.

playbooks/prepare_stack_testconfig.yaml
mdbooth: I copied all this stuff from ocp-doit. We don't actually need it at all and it uses additional VIPs, which is why I moved it into a separate play. The only thing I think this stuff is useful for is if you want to test your openstack by creating VMs manually.

Our team should probably almost never run it. I only didn't delete it because I have used it when testing dev-install.

playbooks/network.yaml
EmilienM: +1 to hardcode

playbooks/prepare_stack.yaml
EmilienM: since we hardcode ``openshift`` in ```playbooks/prepare_stack.yaml``` I would just hardcode it.

playbooks/network.yaml
```diff
@@ -99,41 +61,23 @@
...
+      owner: root
+      group: root
+      mode: '0644'
+    vars:
+      cidr: "{{ hostonly_cidr }}"
+      exit_if: "{{ network_info.public_ipv4.interface }}"
```
EmilienM, line 73: I guess you can remove that one.

playbooks/prepare_stack.yaml
```diff
@@ -27,16 +30,31 @@
...
+          openstack network create --external --share --provider-physical-network external --provider-network-type flat external
       fi
-      if ! openstack subnet show public-subnet; then
-          openstack subnet create public-subnet --subnet-range "{{ public_cidr }}" \
-              --no-dhcp --gateway "{{ public_gateway }}" \
-              --allocation-pool "start={{ public_fip_pool_start }},end={{ public_fip_pool_end }}" \
-              --network public
+      if ! openstack subnet show external-subnet; then
+          openstack subnet create external-subnet --subnet-range "{{ external_cidr }}" \
+              --no-dhcp --gateway "{{ external_gateway }}" \
```
EmilienM, line 40: I think I enabled DHCP to make it work, but maybe in some networks it would work without?

playbooks/prepare_stack.yaml
```diff
@@ -27,16 +30,31 @@
...
+    when: external_fip_pool_start is defined and external_fip_pool_end is defined
+
+  - name: Create hostonly network
+    shell: |
+      if ! openstack network show hostonly; then
+          openstack network create --external --share --provider-physical-network hostonly --provider-network-type flat hostonly
+      fi
+      if ! openstack subnet show hostonly-subnet; then
+          openstack subnet create hostonly-subnet --subnet-range "{{ hostonly_cidr }}" \
+              --no-dhcp --gateway "{{ hostonly_gateway }}" \
```
EmilienM, line 55: same here about dhcp.

EmilienM (changes requested):
@mdbooth I finally was able to test it in my environment, it worked fine modulo a few changes, all commented.

playbooks/prepare_stack.yaml
mdbooth: Done.

playbooks/prepare_stack.yaml
mdbooth: I'm not so sure about this. I guess we'll want to support provider networks both with and without dhcp. I'm going to leave DHCP disabled by default, but add a note to the README on how to enable it (it's a oneliner).

mdbooth: Oh, wait. This is the external network. We definitely can't enable it there! @adduarte would have kittens 😂

playbooks/prepare_stack.yaml
mdbooth: See comment above about enabling this manually.

PR #11, opened 2021-02-17
Enable Octavia

EmilienM:
Enable Octavia (Load Balancer as a Service in OpenStack) by default,
since it's used when enabling Kuryr instead of OpenShiftSDN.

Note that we have to specify both OctaviaGenerateCerts and
OctaviaCaKeyPassphrase, to make the deployment happy but since we don't
use the Amphora driver (we use OVN-provider), it doesn't matter that the
passphrase is not secured.


EmilienM:
This is WIP for now, I'm testing it this week.

mdbooth:
Wow, it would be awesome if it's that simple!

EmilienM:
@mdbooth The last patch is working fine, let me know what you think. Let me know if you don't like putting the keypair generation here but anyway I think we should not care that much about this one.

mdbooth:
LGTM.

PR #15, opened 2021-02-23
e

EmilienM:
_member_ is an old role (OSP16 and before), now it's "member".


mdbooth:
Thanks!

PR #16, opened 2021-03-09
Adds enabled_services as a list

adduarte:
Allows user to customize what services to deploy (tripleo templates)
and  be included as "-e" flags to the "tripleo deploy" command

### DIFF
--- a/playbooks/templates/tripleo-deploy.sh.j2
+++ b/playbooks/templates/tripleo-deploy.sh.j2
@@ -8,7 +8,9 @@ sudo openstack tripleo deploy \
     --local-ip={{ control_plane_ip }} \
     --control-virtual-ip {{ public_api }} \
     -e /usr/share/openstack-tripleo-heat-templates/environments/standalone/standalone-tripleo.yaml \
-    -e /usr/share/openstack-tripleo-heat-templates/environments/services/octavia.yaml \
+{% for service in enabled_services %}
+    -e {{ service }} \
+{% endfor %}
     -r /usr/share/openstack-tripleo-heat-templates/roles/Standalone.yaml \
     -e {{ stack_home }}/containers-prepare-parameters.yaml \
     -e {{ stack_home }}/standalone_parameters.yaml \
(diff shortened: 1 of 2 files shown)


playbooks/templates/tripleo-deploy.sh.j2
```diff
@@ -8,7 +8,9 @@ sudo openstack tripleo deploy \
     --local-ip={{ control_plane_ip }} \
     --control-virtual-ip {{ public_api }} \
     -e /usr/share/openstack-tripleo-heat-templates/environments/standalone/standalone-tripleo.yaml \
-    -e /usr/share/openstack-tripleo-heat-templates/environments/services/octavia.yaml \
+{% for service in enabled_services %}
+    -e {{ service }} \
```
mandre, line 12: If we don't have a need to enable services defined outside of `/usr/share/openstack-tripleo-heat-templates/environments/services/`, can I suggest we do like this:
```suggestion
    -e /usr/share/openstack-tripleo-heat-templates/environments/services/{{ service }}.yaml \
```

This would make the `enabled_services` variable more natural to use.

playbooks/vars/defaults.yaml
```diff
@@ -48,3 +48,6 @@ testconfig_public_key: ~/.ssh/id_rsa.pub
 cirros_url: http://download.cirros-cloud.net/0.5.1/cirros-0.5.1-x86_64-disk.img
 rhcos_meta_url: https://raw.githubusercontent.com/openshift/installer/master/data/data/rhcos.json
 # Define rhcos_url to override use of rhcos_meta_url
+
+enabled_services:
+  - /usr/share/openstack-tripleo-heat-templates/environments/services/octavia.yaml
```
mandre, line 53: If implementing the above suggestion, the `enabled_services` entries would look like:
```suggestion
  - octavia
```

mdbooth:
Agree with @mandre that we'll probably want that list to be tidier if we really start using it. I'm happy to merge this as is and do that when adding more, though. Tidy up patches also welcome.

PR #18, opened 2021-03-09
Changes related to network playbook

adduarte:
- Makes dev-install-network-info.yaml accessible by stack user
         The file is created by the root user (become:yes), thus we need to setup the appropriate ownership/permissions to 
          allow stack user to access the file in following steps. 
- Removed MTUs from network definitions so they are derived from the interfaces added to the bridges.
       If the MTU is not specified by the bridge definition, os-net-config will use the mtus of the attached interfaces.
       Doing so is a better option because it avoids mismatch of mtus. for example, if the interface used for br-ex (eth0) is set to mtu 1400
       we also want the br-ex bridge to have its MTU set to 1400s.

### DIFF
--- a/playbooks/roles/network_info/tasks/main.yaml
+++ b/playbooks/roles/network_info/tasks/main.yaml
@@ -1,7 +1,15 @@
 ---
+- name: Prepare dev-install directory in /var/lib
+  file:
+    path: /var/lib/dev-install
+    state: directory
+    owner: stack
+    group: stack
+    mode: 0775
+
 - name: Read network info if present
   slurp:
-    src: &networkinfo /etc/dev-install-network-info.yaml
+    src: &networkinfo /var/lib/dev-install/dev-install-network-info.yaml
   register: networkinfo_file
   ignore_errors: true
 
@@ -27,4 +35,6 @@
     copy:
       dest: *networkinfo
       content: "{{ network_info | to_nice_json }}"
+      owner: stack
+      group: stack
   when: networkinfo_file.failed
--- a/playbooks/templates/dev-install_net_config.yaml.j2
+++ b/playbooks/templates/dev-install_net_config.yaml.j2
@@ -4,7 +4,6 @@ network_config:
 - type: ovs_bridge
   name: br-ex
   use_dhcp: true
-  mtu: 1500
   ovs_extra:
   - br-set-external-id br-ex bridge-id br-ex
   members:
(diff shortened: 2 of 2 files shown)


playbooks/roles/network_info/tasks/main.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 - name: Read network info if present
   slurp:
-    src: &networkinfo /etc/dev-install-network-info.yaml
+    src: &networkinfo /tmp/dev-install-network-info.yaml
```
mdbooth, line 4: ~~Would prefer `/var/tmp`, as `/tmp` is frequently on tmpfs and not preserved between reboots:~~

~~https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s15.html~~

Although reading that, even `/var/tmp` isn't right. I think we should write this to `/var/lib/`:

https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s08.html

We never want anything to clean it up.

playbooks/templates/dev-install_net_config.yaml.j2
```diff
@@ -4,17 +4,19 @@ network_config:
...
+{% if host_access_interface is defined %}
+    name: {{ host_access_interface }}
+{% else %}
     name: {{ network_info.public_ipv4.interface }}
+{% endif %}
     primary: true
 - type: ovs_bridge
   name: br-ctlplane
   use_dhcp: false
-  mtu: 1500
```
mdbooth, line 17: Yeah, I think I just copied these blindly from somewhere. Definitely safer to remove them.

playbooks/vars/defaults.yaml
```diff
@@ -15,6 +15,11 @@ control_plane_ip: 192.168.24.1
 # By default we use the default IP of the host
 public_api: "{{ network_info.public_ipv4.address }}"
 
+# Interface name which is used to access the host.
+# if not defined it is deduced from network_info.public_ipv4.interface, which will
+# only the first time the network is setup. Subsequent deployments will fail if the interface is not defined here
+# host_access_interface: eth0
```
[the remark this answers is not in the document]
adduarte: taking this out, since I think setting the correct permissions on dev-install-network-info.yaml will solve this indirectlyi

adduarte:
/hold 
this is not working correctly

EmilienM:
Closing it @adduarte we don't need it anymore.

PR #19, opened 2021-03-10
Set the provider network as shared

EmilienM:
So nova can schedule VMs on that network.


mdbooth (approved):
/lgtm

PR #20, opened 2021-03-10
provider network: set dns and enable DHCP

EmilienM:
Or instances don't get an IP and can't resolve domain names from
Internet.

### DIFF
--- a/playbooks/prepare_stack.yaml
+++ b/playbooks/prepare_stack.yaml
@@ -97,7 +97,8 @@
       fi
       if ! openstack subnet show hostonly-subnet; then
           openstack subnet create --project openshift hostonly-subnet --subnet-range "{{ hostonly_cidr }}" \
-              --no-dhcp --gateway "{{ hostonly_gateway }}" \
+              --dhcp --gateway "{{ hostonly_gateway }}" \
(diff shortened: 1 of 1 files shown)


playbooks/prepare_stack.yaml
```diff
@@ -97,7 +97,8 @@
       fi
       if ! openstack subnet show hostonly-subnet; then
           openstack subnet create --project openshift hostonly-subnet --subnet-range "{{ hostonly_cidr }}" \
-              --no-dhcp --gateway "{{ hostonly_gateway }}" \
+              --dhcp --gateway "{{ hostonly_gateway }}" \
+              --dns-nameserver "{{ network_info.dns | first }}" \
```
mdbooth, line 101:
```
  --dns-nameserver <dns-nameserver>
                        DNS server for this subnet (repeat option to set multiple DNS servers)
```

Is it worth repeating this for all nameservers?

mdbooth (approved):
/lgtm

Was cautious of adding dhcp to `external`, but hostonly makes sense. Suggestion below.

playbooks/prepare_stack.yaml
mdbooth: Depends how much effort it is. Nice to have, hence I was happy to merge without it.

PR #21, opened 2021-03-19
Improvements to support other OSP versions

EmilienM:
- Change location of dev-install-network-info.yaml
- Templatize ContainerImagePrepare
- Allow to override net-config in 16

### DIFF
--- a/playbooks/templates/standalone_parameters.yaml.j2
+++ b/playbooks/templates/standalone_parameters.yaml.j2
@@ -21,7 +21,14 @@ parameter_defaults:
   StandaloneHomeDir: "{{ ansible_env.HOME }}"
   InterfaceLocalMtu: 1500
   SELinuxMode: permissive
+  # For OSP17+:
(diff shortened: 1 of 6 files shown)


playbooks/templates/standalone_parameters.yaml.j2
```diff
@@ -22,6 +22,11 @@ parameter_defaults:
   InterfaceLocalMtu: 1500
   SELinuxMode: permissive
   StandaloneNetworkConfigTemplate: "{{ ansible_env.HOME }}/dev-install_net_config.yaml"
+  StandaloneNetConfigOverride:
```
mdbooth, line 25: I might put a comment or even an if block here (to entirely omit it) to make it really obvious that this is OSP16 only. I wasted nearly 2 days working this out first time! Wouldn't want anybody to do that again just because we're setting it in this file.

mdbooth:
This looks really good. Let me know when you're done testing.

PR #31, opened 2021-04-19
Move to t-o-a & support centos/rdo

EmilienM:
- Install local requirements to use TripleO Ansible Operators
- CentOS / RDO support

### DIFF
--- a/example-overrides/local-overrides-centos8-tripleo-train.yaml
+++ b/example-overrides/local-overrides-centos8-tripleo-train.yaml
@@ -1,0 +1,10 @@
+standalone_host: <standalone FQDN>
+# Both public_api and control_plane_ip have to be the same in OSP16 for Pacemaker
+public_api: <IP address used to reach the node>
+control_plane_ip: <Same IP address used to reach the node>
+tripleo_repos_branch: train
+cip_config:
+  - set:
+      namespace: quay.io/tripleotrain
+      name_prefix: openstack-
+      tag: current-tripleo
--- a/playbooks/prepare_host.yaml
+++ b/playbooks/prepare_host.yaml
@@ -2,20 +2,42 @@
 - hosts: standalone
   become: true
   become_user: root
-  gather_facts: false
+  gather_facts: true
   vars_files: vars/defaults.yaml
 
   tasks:
 
-  - name: install rhos-release
-    yum:
-      state: installed
-      name:
-      - http://download.devel.redhat.com/rcm-guest/puddles/OpenStack/rhos-release/rhos-release-latest.noarch.rpm
-      disable_gpg_check: True
+  - name: Prepare host on RHEL system
+    when:
+      - ansible_facts.distribution == 'RedHat'
+    block:
+      - name: install rhos-release
+        yum:
+          state: installed
+          name:
+          - http://download.devel.redhat.com/rcm-guest/puddles/OpenStack/rhos-release/rhos-release-latest.noarch.rpm
+          disable_gpg_check: True
 
-  - name: Configure rhos release "{{ rhos_release }}"
-    shell: rhos-release "{{ rhos_release }}"
+      - name: Configure rhos release "{{ rhos_release }}"
+        shell: rhos-release "{{ rhos_release }}"
+
+      - name: Fetch the Red Hat root certificate
+        get_url:
+          dest: /etc/pki/ca-trust/source/anchors/RH-IT-Root-CA.crt
+          url: https://password.corp.redhat.com/RH-IT-Root-CA.crt
+
+      - name: Add the certificate to the local trust bundle
+        shell: |
+          update-ca-trust enable
+          update-ca-trust extract
+
+  - name: Prepare host on CentOS system
+    when:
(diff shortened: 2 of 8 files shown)


playbooks/prepare_host.yaml
```diff
@@ -2,20 +2,42 @@
...
+        yum:
+          state: installed
+          name:
+          - http://download.devel.redhat.com/rcm-guest/puddles/OpenStack/rhos-release/rhos-release-latest.noarch.rpm
+          disable_gpg_check: True
+
+      - name: Configure rhos release "{{ rhos_release }}"
+        shell: rhos-release "{{ rhos_release }}"
+
+      - name: Fetch the Red Hat root certificate
```
mdbooth, line 24: This is likely something we want regardless of whether we're deploying OSP or RDO. IIRC we need this to pull from internal registries when installing openshift.

mdbooth, line 24: I think we also need this when pulling directly from internal registries.

However... that's dependent on where we run the installer from, so I'm ok with ditching this here and adding it back anywhere else it's needed when we find it, perhaps after refactoring it into a role.

EmilienM: let's keep it specific to RHEL now and I'm happy to fix it later if we hit the issue.

example-overrides/local-overrides-centos8-tripleo-train.yaml
```diff
@@ -0,0 +1,10 @@
+standalone_host: <standalone FQDN>
+# Both public_api and control_plane_ip have to be the same in OSP16 for Pacemaker
+public_api: <IP address used to reach the node>
+control_plane_ip: <Same IP address used to reach the node>
+tripleo_repos_branch: train
+cip_config:
```
mdbooth, line 6: Is this used automatically by something?

EmilienM: this is something I'll refactor later to use the tripleo operator for container image prepare role from there. For now, this is just an example on how to override the default we have in dev-install, and yes this is used as an environment input for when deploying standalone.

example-overrides/local-overrides-centos8-tripleo-train.yaml
```diff
@@ -0,0 +1,10 @@
+standalone_host: <standalone FQDN>
+# Both public_api and control_plane_ip have to be the same in OSP16 for Pacemaker
+public_api: <IP address used to reach the node>
+control_plane_ip: <Same IP address used to reach the node>
```
mdbooth, line 4: I don't think this is correct. I'm not entirely surely sure how control_plane_ip is used by OOO, tbh. The big one afaik is:

```
sudo openstack tripleo deploy \
    --templates \
```

I have never looked under the covers at what --local-ip actually sets, but I do know that we add it to br-ctlplane. I don't think we want the same IP added to both br-ctlplane and br-external. The standalone example used 192.168.24.1. Do you have a specific reason for wanting to change this?

EmilienM: Yeah we discussed about it (btw I copy pasted from the OSP16 local overrides example), and this is something we need to investigate further.
This has to do with Pacemaker deploying in Train, I don't remember all the details yet.

example-overrides/local-overrides-centos8-tripleo-train.yaml
mdbooth: Ok, so it wouldn't be used by anything in this commit?

EmilienM: this is an overrides example if you plan to deploy on centos/RDO, train version (upstream for OSP16), so yes this should be used.

example-overrides/local-overrides-centos8-tripleo-train.yaml
mdbooth: As I say, I didn't really understand this stuff when I added it.

It sounds like we should remove control_plane_ip, which suggests we can also remove br-ctlplane. We'd then define deploy with local-ip and control-virtual-ip both using `public_api`?

example-overrides/local-overrides-centos8-tripleo-train.yaml
mdbooth: It's used by `tripleo.operator.tripleo_repos`?

example-overrides/local-overrides-centos8-tripleo-train.yaml
EmilienM: I think this isn't related to this PR, I just provided an example of an override based on OSP16 but for upstream deployment.
Yes we should fix that problem, but not in this PR imo.
Note that we don't need this hack on master.

example-overrides/local-overrides-centos8-tripleo-train.yaml
EmilienM: no, it creates an environment file that is used by standalone.

EmilienM:
@mdbooth I fixed conflict and this is ready to go I think. Let me know if you need any change to happen.

PR #32, opened 2021-04-19
Enable advanced nic configs (e.g. SR-IOV)

EmilienM:
Allow the user to override the network config when they know what
they're doing, if not we use the opinionated template like before.

### DIFF
--- a/playbooks/templates/dev-install_net_config.yaml.j2
+++ b/playbooks/templates/dev-install_net_config.yaml.j2
@@ -1,6 +1,9 @@
 #vi:syntax=yaml
 ---
 network_config:
+{% if network_config is defined %}
+{{ network_config | to_nice_yaml }}
+{% else %}
 - type: ovs_bridge
   name: br-ex
   use_dhcp: true
@@ -38,3 +41,4 @@ network_config:
   - type: interface
     name: dummy1
     nm_controlled: true
+{% endif %}
--- a/playbooks/vars/defaults.yaml
+++ b/playbooks/vars/defaults.yaml
@@ -20,6 +20,10 @@ control_plane_ip: 192.168.24.1
 # By default we use the default IP of the host
 public_api: "{{ network_info.public_ipv4.address }}"
 
+# For advanced network configurations (e.g. SR-IOV), network_config can be overriden
+# The format is for os-net-config, check dev-install_net_config.yaml.j2 template.
+# network_config
+
 # The `external` provider network
 # This will be created only if external_fip_pool_start and external_fip_pool_end
 # are manually defined.


playbooks/templates/dev-install_net_config.yaml.j2
```diff
@@ -1,6 +1,9 @@
 #vi:syntax=yaml
 ---
 network_config:
+{% if network_config is defined %}
+{{ network_config | to_nice_yaml }}
```
adduarte, line 5: This would actually be better named "network_config_extra" or something. 
because what is really happening is that whatever the variable "network_config" has in default.yaml is being added to the network_config. Or is is replacing it?

playbooks/vars/defaults.yaml
```diff
@@ -20,6 +20,10 @@ control_plane_ip: 192.168.24.1
 # By default we use the default IP of the host
 public_api: "{{ network_info.public_ipv4.address }}"
 
+# For advanced network configurations (e.g. SR-IOV), network_config can be overriden
+# The format is for os-net-config, check dev-install_net_config.yaml.j2 template.
+# network_config
+
```
adduarte, line 26: what if instead of completely removing the network_config, you add to it. 
reason being that then the user does not need to worry about adding the dummy interface or what not. 
Or maybe copy and paste the br-ex and dummy interface definitions to here as an example to be added for a full functioning network_config.

EmilienM: There are multiple ways to solve this problem:

- we can add configs to the default but then someone will ask to remove stuffs from the default
- we can completely override the default
- ...

At the end it ends up a bit complex in my opinion, and needs to be maintained.
I think that if you need a network config that isn't the default, you're advanced enough to provide the whole config.

I'm happy to implement whatever solution, but I think we need to keep it stupid and simple and not try to cover all cases.

mdbooth (commented):
I'm not a fan of this because it's not simple to use. I think advanced users can just use TripleO. My 2 main goals for dev-install are:

1. Have an extremely low barrier to entry.

The user just wants an OpenStack ready to go. They don't want to learn TripleO to achieve that, and the prospect of having to learn TripleO is likely to dissuade people from even starting.

2. When somebody needs to learn some TripleO to do something new, I want to capture that in dev-install so that nobody else has to learn it.

My worry is that by adding a complete network config override we might discourage the team from improving the tool.

With SR-IOV specifically my hope was that we could do something like @adduarte described. I imagined that we could, for e.g. define an sriov_interfaces variable which is an empty list by default. Then we can loop over them in the right place in the os-net-config.

Another idea I thought we might steal from OpenStack CI is the concept of a scenario/topology. This is for when you have to concede that you can't do everything in a single opinionated configuration because some things are mutually exclusive (e.g. testing both with and without feature X). We could maintain N extremely opinionated configurations, but we would want to minimise N to the greatest possible degree.

I want to maintain dev-install's focus on being a tool for very simply deploying a fully functional but disposable OpenStack.

With all that said, if I haven't persuaded you I'm also ok if you want to merge this: it doesn't actually prevent anything after all.

mdbooth:
I discussed this with @EmilienM . I can see there may be occasions where it could be useful, but I will be disappointed to see it used a lot.

PR #35, opened 2021-04-23
WIP DO NOT MERGE only for example: Changes required for SRIOV

adduarte:
- Changes to the GRUB_CMDLINE_LINUX need to be done and the host rebooted.
   The changes include hugpage support and other required kernel parameters.

### DIFF
--- a/playbooks/install_stack.yaml
+++ b/playbooks/install_stack.yaml
@@ -62,6 +62,11 @@
       src: containers-prepare-parameters.yaml.j2
       dest: "{{ ansible_env.HOME }}/containers-prepare-parameters.yaml"
 
+  - name: Create StandaloneSriov Role
+    shell: |
+      cp /usr/share/openstack-tripleo-heat-templates/roles/Standalone.yaml {{ ansible_env.HOME }}/Standalone.yaml
+      echo "    - OS::TripleO::Services::NeutronSriovAgent" >> $HOME/Standalone.yaml
+
   # NOTE(mdbooth): The deploy step consistently fails to return. It fails in a
   # manner consistent with a hung SSH connection. The following dance is an
   # effort to never have a very long-running ansible task which might hang.
--- a/playbooks/templates/dev-install_net_config.yaml.j2
+++ b/playbooks/templates/dev-install_net_config.yaml.j2
@@ -38,3 +38,11 @@ network_config:
   - type: interface
     name: dummy1
     nm_controlled: true
+- type: sriov_pf
+  name: {{ sriov_interface }}
+  numvfs: 10
+  use_dhcp: false
+  defroute: false
+  nm_controlled: true
+  hotplug: true
+  promisc: false
--- a/playbooks/templates/standalone_parameters.yaml.j2
+++ b/playbooks/templates/standalone_parameters.yaml.j2
@@ -17,6 +17,8 @@ parameter_defaults:
   NeutronDnsDomain: {{ clouddomain }}
   NeutronBridgeMappings: {{ neutron_bridge_mappings }}
   NeutronFlatNetworks: {{ neutron_flat_networks }}
+  NeutronNetworkType: 'flat,vlan'
+  NeutronNetworkVLANRanges: 'sriov:100:199'
   StandaloneEnableRoutedNetworks: false
   StandaloneHomeDir: "{{ ansible_env.HOME }}"
   InterfaceLocalMtu: 1500
@@ -33,3 +35,18 @@ parameter_defaults:
   OctaviaCaKeyPassphrase: "secrete"
   OctaviaAmphoraSshKeyFile: "{{ ansible_env.HOME }}/octavia.pub"
   OctaviaAmphoraImageFilename: "{{ ansible_env.HOME }}/amphora.qcow2"
+
+  NeutronPhysicalDevMappings: "{{ neutron_dev_physical_mappings }}"
+  NovaReservedHostMemory: 4096
+
+# The following kernelargs, need to be added to the host GRUB_CMDLINE_LINUX
+# and the host needs to be rebooted and the values cheked.
+#  #KernelArgs: "default_hugepagesz=1GB hugepagesz=1G hugepages=32 iommu=pt intel_iommu=on isolcpus=1-19,21-39"
+#  Still don't hav TunedProfile working correctly
+#  TunedProfileName: "cpu-partitioning"
+# The PCI passthrough access address should be provided by user as variable
+  NovaPCIPassthrough:
+    - address: "0000:04:00.1"
+      trusted: "true"
+      physical_network: "sriov"
+  NeutronEnableDVR: false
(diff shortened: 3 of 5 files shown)


playbooks/install_stack.yaml
```diff
@@ -62,6 +62,11 @@
       src: containers-prepare-parameters.yaml.j2
       dest: "{{ ansible_env.HOME }}/containers-prepare-parameters.yaml"
 
+  - name: Create StandaloneSriov Role
+    shell: |
+      cp /usr/share/openstack-tripleo-heat-templates/roles/Standalone.yaml {{ ansible_env.HOME }}/Standalone.yaml
+      echo "    - OS::TripleO::Services::NeutronSriovAgent" >> $HOME/Standalone.yaml
+
```
adduarte, line 69: This is used to create the Standalone SRIOV role. 
Tried using StandaloneSRIOV.yaml as the name but there were too many places where Standalone would need to be used so just left as same name. Probably can be named better with more investigation

playbooks/templates/dev-install_net_config.yaml.j2
```diff
@@ -38,3 +38,11 @@ network_config:
   - type: interface
     name: dummy1
     nm_controlled: true
+- type: sriov_pf
+  name: {{ sriov_interface }}
+  numvfs: 10
```
adduarte, line 43: numvfs should be a variable in defaults.yaml

playbooks/templates/standalone_parameters.yaml.j2
```diff
@@ -33,3 +35,18 @@ parameter_defaults:
...
   OctaviaAmphoraImageFilename: "{{ ansible_env.HOME }}/amphora.qcow2"
+
+  NeutronPhysicalDevMappings: "{{ neutron_dev_physical_mappings }}"
+  NovaReservedHostMemory: 4096
+
+# The following kernelargs, need to be added to the host GRUB_CMDLINE_LINUX
+# and the host needs to be rebooted and the values cheked.
+#  #KernelArgs: "default_hugepagesz=1GB hugepagesz=1G hugepages=32 iommu=pt intel_iommu=on isolcpus=1-19,21-39"
+#  Still don't hav TunedProfile working correctly
+#  TunedProfileName: "cpu-partitioning"
```
adduarte, line 46: when this is left uncommented, the installation errors out wth a message about not finding the TuneProfile cpu-partitioning.
This will affect the cpu allocation to the vm thus is needed by dpdk

playbooks/templates/standalone_parameters.yaml.j2
```diff
@@ -33,3 +35,18 @@ parameter_defaults:
...
+# The PCI passthrough access address should be provided by user as variable
+  NovaPCIPassthrough:
+    - address: "0000:04:00.1"
```
adduarte, line 49: The address needs to be passed in as a variable. The value is obtained by looking at which sr-iov supporthing nic will be used under the the system menu. But user should provide it.

playbooks/templates/tripleo-deploy.sh.j2
```diff
@@ -7,10 +7,10 @@ sudo openstack tripleo deploy \
     --templates \
     --local-ip={{ control_plane_ip }} \
     --control-virtual-ip {{ public_api }} \
-    -e /usr/share/openstack-tripleo-heat-templates/environments/standalone/standalone-tripleo.yaml \
-{% for service in enabled_services %}
+  {% for service in enabled_services %}
     -e {{ service }} \
-{% endfor %}
+  {% endfor %}
+    -e /usr/share/openstack-tripleo-heat-templates/environments/standalone/standalone-tripleo.yaml \
```
adduarte, line 13: Needed to move the enabled services up (not sure if absolutely necessary, but it seemed it would not work correctly the other way around. could have been user error.

playbooks/vars/defaults.yaml
```diff
@@ -56,8 +56,11 @@ rhcos_meta_url: https://raw.githubusercontent.com/openshift/installer/master/dat
 # Define rhcos_url to override use of rhcos_meta_url
 
 enabled_services:
-  - /usr/share/openstack-tripleo-heat-templates/environments/services/octavia.yaml
-standalone_role: /usr/share/openstack-tripleo-heat-templates/roles/Standalone.yaml
+  - /usr/share/openstack-tripleo-heat-templates/environments/services/neutron-ovn-sriov.yaml
+#  - /usr/share/openstack-tripleo-heat-templates/environments/services/octavia.yaml
```
adduarte, line 60: octavia probably does not need to be disabled. Just did for this run for speed. and clarity.

adduarte (commented):
just to share example and experience

playbooks/install_stack.yaml
EmilienM: For this one, I'll see if we can add OS::TripleO::Services::NeutronSriovAgent into the upstream Standalone role. It's disabled by default anyway:

```
overcloud-resource-registry-puppet.j2.yaml:
166:  OS::TripleO::Services::NeutronSriovAgent: OS::Heat::None
```

playbooks/templates/dev-install_net_config.yaml.j2
EmilienM: so I think we should use what we did in https://github.com/shiftstack/dev-install/pull/32/commits/6a33b009b400508fa58f827ed892acd490178f6a to override the network_config when needed.
Or simply do what I think @mdbooth would prefer:

```
- type: sriov_pf
  name: {{ sriov_interface }}
  numvfs: {{ sriov_nic_numvfs | defauly(10) }}
  use_dhcp: false
  defroute: false
  nm_controlled: true
  hotplug: true
  promisc: false
{% endif %}
```

I have no strong opinion, @adduarte if you think this config ^ can work in a majority of SRIOV configs, then I think this is good for dev-install.

playbooks/templates/standalone_parameters.yaml.j2
```diff
@@ -17,6 +17,8 @@ parameter_defaults:
   NeutronDnsDomain: {{ clouddomain }}
   NeutronBridgeMappings: {{ neutron_bridge_mappings }}
   NeutronFlatNetworks: {{ neutron_flat_networks }}
+  NeutronNetworkType: 'flat,vlan'
```
EmilienM, line 20: I think the default in TripleO with OVN is ['geneve', 'vxlan', 'vlan', 'flat'].
We should move that to ```playbooks/vars/defaults.yaml``` instead.

playbooks/vars/defaults.yaml
EmilienM: yeah that is something we can do via the overrides, maybe in your PR you could add an example of overrides in https://github.com/shiftstack/dev-install/tree/master/example-overrides

playbooks/templates/tripleo-deploy.sh.j2
EmilienM: this was not needed in my testing, btw.

EmilienM:
Closing it @adduarte we don't need it anymore.

PR #37, opened 2021-04-23
Add new parameter "extra_heat_params"

EmilienM:
extra_heat_params is a new parameter that'll let you set or override any
Heat parameter.

e.g.

```
extra_heat_params:
      NovaReservedHostMemory: 4096
      NovaPCIPassthrough:
        - address: "0000:04:00.1"
```

Note that if a Heat param is defined in extra_heat_params and also in standalone_parameters,
the former will override the latter which can be useful if you need specific configs.


mdbooth (approved):
LGTM.

PR #38, opened 2021-04-26
Add support to install Ceph via loop device

EmilienM:
For the developer use case, we use a loop device to install LVM volumes
for ceph.
This patch will add two options to dev-install:

* ceph_enabled (false by default)
* ceph_loop_device_size (10 GB by default)

It will take care of creating a loop device that is reboot-persistent,
and then use it for the deployment of Ceph via ceph-ansible.
It'll also dynamically load the environment and services if we want
Ceph enabled.

This addition does not configure Ceph for production, so in the
OpenShift on OpenStack use-case, keep in mind that the Ceph
performances won't be great on loop devices. In other words, booting
the cluster nodes from volumes will be problematic because of perf
requirements for etcd.

It also adds nodisk flavors for both large and xlarge. They can be
useful when deploying OCP masters or workers with rootVolume from
Cinder (backend in Ceph).

### DIFF
--- a/playbooks/prepare_host.yaml
+++ b/playbooks/prepare_host.yaml
@@ -69,3 +69,39 @@
     user:
       name: stack
       groups: wheel
+  - name: Prepare host for Ceph
+    when: ceph_enabled
+    block:
+    - name: Make sure we have losetup installed/latest
+      package:
+        name:
+          - util-linux
+          - lvm2
+        state: latest
+    - name: Use dd and losetup to create the loop devices
+      shell: |
+        fallocate -l {{ ceph_loop_device_size }}G /var/lib/ceph-osd.img
+        losetup /dev/ceph-loopback /var/lib/ceph-osd.img
+        pvcreate /dev/ceph-loopback
+        vgcreate vg2 /dev/ceph-loopback
+        lvcreate -n data-lv2 --size {{ ceph_loop_device_size * 96 / 100 }}G vg2
+        lvcreate -n db-lv2 --size {{ (ceph_loop_device_size * 4 / 100) - 0.1 }}G vg2
+    - name: Create /etc/systemd/system/ceph-osd-losetup.service
+      copy:
+        dest: /etc/systemd/system/ceph-osd-losetup.service
+        content: |
+          [Unit]
+          Description=Ceph OSD losetup
+          After=syslog.target
+          [Service]
+          Type=oneshot
+          ExecStart=/bin/bash -c '/sbin/losetup /dev/ceph-loopback || \
+          /sbin/losetup /dev/ceph-loopback /var/lib/ceph-osd.img ; partprobe /dev/ceph-loopback'
+          ExecStop=/sbin/losetup -d /dev/ceph-loopback
+          RemainAfterExit=yes
+          [Install]
+          WantedBy=multi-user.target
+    - name: Enable ceph-osd-losetup.service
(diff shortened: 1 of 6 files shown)


playbooks/prepare_host.yaml
```diff
@@ -69,3 +69,39 @@
...
+    block:
+    - name: Make sure we have losetup installed/latest
+      package:
+        name:
+          - util-linux
+          - lvm2
+        state: latest
+    - name: Use dd and losetup to create the loop devices
+      shell: |
+        dd if=/dev/zero of=/var/lib/ceph-osd.img bs=1 count=0 seek={{ ceph_loop_device_size }}G
```
mdbooth, line 83: Suggestion:

```
truncate -s {{ ceph_loop_device_size}}G /var/lib/ceph-osd.img
```

This will create a sparse file, so it can be as big as you like as long as you don't actually try to use all of it 😀 If we decided not to use sparse:

```
fallocate -l {{ ceph_loop_device_size * <make it bytes> }} /var/lib/ceph-osd.img
```

This will preallocate the space and do lazy zeroing if the filesystem supports it (which it will), so it's very fast.

playbooks/prepare_host.yaml
```diff
@@ -69,3 +69,39 @@
...
+        losetup /dev/loop3 /var/lib/ceph-osd.img
```
mdbooth, line 84: If we're naming this loop device explicitly, might be better to be clearer and avoid potential conflicts by calling it something like `/dev/ceph-loopback`?

playbooks/prepare_host.yaml
```diff
@@ -69,3 +69,39 @@
...
+      copy:
+        dest: /etc/systemd/system/ceph-osd-losetup.service
+        content: |
+          [Unit]
+          Description=Ceph OSD losetup
+          After=syslog.target
+          [Service]
+          Type=oneshot
+          ExecStart=/bin/bash -c '/sbin/losetup /dev/loop3 || \
+          /sbin/losetup /dev/loop3 /var/lib/ceph-osd.img ; partprobe /dev/loop3'
```
mdbooth, line 99: I think the partprobe is redundant here because there's no partition table on it.

playbooks/vars/defaults.yaml
```diff
@@ -69,3 +69,8 @@ neutron_flat_networks: "external,hostonly"
 tripleo_repos_repos:
   - current-tripleo-dev
   - ceph
+
+ceph_enabled: false
```
mdbooth, line 73: This is the kind of config complexity I like 😁

Any idea how much time/resource it adds to the installation? IOW, any reason not to make the default true?

mdbooth (approved):
Looks great. Some suggestions inline, but happy to merge with or without.

playbooks/prepare_host.yaml
EmilienM: I'll leave it here, I stole it from tripleo/ceph CI, I'm sure there is a reason behind this, maybe it's for the LVM partitions?

playbooks/vars/defaults.yaml
EmilienM: it's adding a good 15min and it's a loop device, so perfs aren't great, I wouldn't make it the default for now.

EmilienM:
I addressed Matt's comments, i'll merge it.

PR #42, opened 2021-04-27
Enable SR-IOV + DPDK support

EmilienM:
* Add an interface to easily add new services into the default Standalone role.
* Make it so SR-IOV and DPDK configs are independently configurable
* Configure SR-IOV + DPDK parameters in standalone_parameters.
* Configure the SR-IOV NIC.
* Modify the Standalone role with SR-IOV services
* Reboot the node at the end of the deployment to apply Kernel changes.

### DIFF
--- a/playbooks/templates/dev-install_net_config.yaml.j2
+++ b/playbooks/templates/dev-install_net_config.yaml.j2
@@ -38,7 +38,18 @@ network_config:
   - destination: {{ hostonly_cidr }}
     nexthop: {{ hostonly_gateway }}
   members:
+{% if sriov_interface is defined %}
+  - type: sriov_pf
+    name: {{ sriov_interface }}
+    numvfs: {{ sriov_nic_numvfs | mandatory }}
+    use_dhcp: false
+    defroute: false
+    nm_controlled: false
+    hotplug: true
+    promisc: false
+{% else %}
   - type: interface
     name: dummy1
     nm_controlled: true
+{% endif %}
 {% endif %}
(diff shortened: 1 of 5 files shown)


playbooks/vars/defaults.yaml
```diff
@@ -100,3 +103,15 @@ ceph_enabled: false
 # Size of the loop device that will be
 # used for Ceph (in GB).
 ceph_loop_device_size: 10
+
+sriov_enabled: false
```
adduarte, line 107: nit. is this needed?
can it be inferred from sriov_services being defined on any of the others settings

playbooks/templates/dev-install_net_config.yaml.j2
```diff
@@ -42,3 +42,13 @@ network_config:
     name: dummy1
     nm_controlled: true
 {% endif %}
+{% if sriov_enabled %}
+- type: sriov_pf
+  name: {{ sriov_interface | mandatory }}
```
adduarte, line 47: here, sriov_enabled = true could be inferred  from the presence of srivo_interface being defined.

adduarte (commented):
its looking good. I don't see any major changes needed

playbooks/templates/dev-install_net_config.yaml.j2
EmilienM: yeah I'll do that instead.

EmilienM:
@mdbooth @mandre @adduarte 
This PR has been tested by myself and deployed a cloud with SR-IOV / DPDK, and I ran some workloads which seem to work. I think we can go ahead with this first iteration.

Thanks

mandre (approved):

PR #50, opened 2021-05-06
A mix of QoL and bug fixes

mdbooth:
Reduce required configuration:
* Remove control_plane_ip
* Configure cip_config automatically from puddle
* Enable ceph by default (but you can still disable it)
* Allow SR-IOV without DPDK

Fixes:
* Idempotent ceph vg creation means you can re-run prepare_host
* Fix network systemd unit by disabling unused interfaces

### DIFF
--- a/playbooks/install_stack.yaml
+++ b/playbooks/install_stack.yaml
@@ -81,11 +81,34 @@
       url: https://images.rdoproject.org/octavia/master/amphora-x64-haproxy-centos.qcow2
       dest: "{{ ansible_env.HOME }}/amphora.qcow2"
 
+  - name: Set cip_config from downloaded container_image_prepare.yaml
+    when: cip_config is not defined
+    block:
+    - name: Read container_image_prepare.yaml
+      slurp:
+        src: /home/stack/container_image_prepare.yaml
+      register: cip_yaml
+    - set_fact:
+        cip_raw: "{{ (cip_yaml.content | b64decode | from_yaml)['container-image-prepare'] | dict2items }}"
+    - set_fact:
+        cip_config: "{{ [{ 'set': dict(keys | zip(values)) }] }}"
+      vars:
+        # container_image_prepare.yaml downloaded from the puddle contains some
+        # keys with invalid names, e.g. 'ceph-namespace' instead of
+        # 'ceph_namespace'. We rewrite them.
+        keys: "{{ cip_raw | map(attribute='key') | map('regex_replace', '-', '_') | list }}"
+        values: "{{ cip_raw | map(attribute='value') | list }}"
+
   - name: Create containers-prepare-parameters.yaml
-    template:
-      mode: 0644
-      src: containers-prepare-parameters.yaml.j2
+    copy:
       dest: "{{ ansible_env.HOME }}/containers-prepare-parameters.yaml"
+      content: "{{ cip_content | to_nice_yaml }}"
+      owner: stack
+      mode: 0644
+    vars:
+      cip_content:
+        parameter_defaults:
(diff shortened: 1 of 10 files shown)


playbooks/install_stack.yaml
```diff
@@ -131,7 +154,7 @@
       tripleo_deploy_standalone: true
       tripleo_deploy_output_dir: "{{ ansible_env.HOME }}"
       tripleo_deploy_local_ip: "{{ public_api }}"
-      tripleo_deploy_control_virtual_ip: "{{ control_plane_ip }}"
+      tripleo_deploy_control_virtual_ip: "{{ public_api }}"
```
EmilienM, line 157: this is something that I need to test against tripleo master because I'm not sure it works the same way between train & master for some reasons that I never understood honestly.

playbooks/prepare_host.yaml
```diff
@@ -106,3 +113,18 @@
...
+    - name: Create ceph data LV
+      lvol:
+        vg: vg_ceph
+        lv: data
+        size: 96%VG
+    - name: Create ceph db LV
+      lvol:
+        vg: vg_ceph
+        lv: db
+        size: 4%VG
```
EmilienM, line 130: I don't think this will work because some space is lost for metadatas, which is why I did `- 0.1`

mdbooth: It seems to work 🤷‍♂️ I'd hold off changing it until it actually fails.

playbooks/vars/defaults.yaml
```diff
@@ -95,10 +95,10 @@ tripleo_repos_repos:
   - current-tripleo-dev
   - ceph
 
-ceph_enabled: false
+ceph_enabled: true
```
EmilienM, line 77: this is too opinionated, IMO.

- in real life, people won't use loop devices to test OCP clusters in VMs backed by Ceph loop devices, the cluster won't start it's too slow
- not everyone has 100GB of available

This feature is useful to test things like "hey can I create a volume" or testing Manila/Ceph integration.
But IMO we shouldn't enable this by default.

playbooks/prepare_host.yaml
```diff
@@ -31,6 +33,20 @@
           update-ca-trust enable
           update-ca-trust extract
 
+      - name: Configure rhos release "{{ rhos_release }}" and keep puddle name
+        shell: rhos-release "{{ rhos_release }}" | awk '/^# rhos-release/ { print $5 }'
```
EmilienM, line 37: really cool!!

playbooks/prepare_host.yaml
EmilienM: ack

playbooks/vars/defaults.yaml
mdbooth: I strongly disagree about this 😉

Most devs will want ceph and won't want to reinstall because they forgot to enable it, and anybody using a DSAL box has way more than 100G disk going spare.

I think disabling ceph is an interesting, but niche configuration. I'm happy for the niche configuration to require additional configuration. I don't want everybody else to require it.

EmilienM: **All** DSAL users have > 1 disk and will dedicate a disk (e.g. */dev/sdb*) to create the LVM configuration and have good performances using direct block storage, vs using a loop device which we all know doesn't provide great I/O results.

EmilienM: If you merges https://github.com/shiftstack/dev-install/pull/52, i'm fine with this change.

mdbooth:
I've merged all the parts of this PR which I want to separately.

PR #52, opened 2021-05-06
If ceph is enable, disable Nova RBD

EmilienM:
Since we use a loop device for the Ceph disk, performances wouldn't be
great enough to boot VMs (e.g. OCP cluster) and would lead to problems
later.
If anyone wants wants to overrides this option, they can do it via:

```
extra_heat_params:
  NovaEnableRbdBackend: true
```
To use at your own risks...


mdbooth (approved):
Makes sense to me. It makes no difference to us how Nova stores its data, and local disk is going to be faster.

PR #55, opened 2021-05-07
Add manila_enabled parameter

mdbooth:


playbooks/install_stack.yaml
```diff
@@ -70,6 +70,29 @@
...
+    - name: Add manila extraconfig
+      set_fact:
+        standalone_extra_config: "{{ standalone_extra_config | combine(manila_extra_config) }}"
+      vars:
+        manila_extra_config:
+          ganesha_vip: "{{ public_api }}"
+
+    - name: Ensure ceph is enabled
+      set_fact:
+        ceph_enabled: true
```
EmilienM, line 94: Ceph users who don't use loop devices won't be able to use Manila because they don't use that option (otherwise they get Ceph configured with a loop device).

The simplest option to me is to add a toggle to whether or not create the loop device, and use it where it's useful (in prepare_host when we create the loop device) and in standalone_parameters when configuring Ceph params.

playbooks/templates/standalone_parameters.yaml.j2
```diff
@@ -69,6 +69,12 @@ parameter_defaults:
   CephPoolDefaultPgNum: 8
   CephPoolDefaultSize: 1
 {% endif %}
+{% if standalone_extra_config|length > 0 %}
+  StandaloneExtraConfig: {{ standalone_extra_config }}
+  {% for key, value in standalone_extra_config.iteritems() %}
```
mandre, line 74: This is python2 syntax and fails in the `Create standalone_parameters.yaml` task:

```
TASK [Create standalone_parameters.yaml] *********************************************************************************************************************************
fatal: [standalone]: FAILED! => {"changed": false, "msg": "AnsibleUndefinedVariable: 'dict object' has no attribute 'iteritems'"}                                         
```


```suggestion
  {% for key, value in standalone_extra_config.items() %}
```

mdbooth (commented):
/lgtm

@EmilienM added some additional magic sauce for his Vexxhost box:

```
  CephPoolDefaultPgNum: 32
  CephPoolDefaultSize: 1
```

It wasn't clear to me whether we also need to add that here?

EmilienM:
> /lgtm
> 
> @EmilienM added some additional magic sauce for his Vexxhost box:
> 
> ```
>   CephPoolDefaultPgNum: 32
>   CephPoolDefaultSize: 1
> ```
> 
> It wasn't clear to me whether we also need to add that here?

32 is indeed better as it allows more PG to be created but I think 8 was enough for my testing...

mandre:
> > /lgtm
> > @EmilienM added some additional magic sauce for his Vexxhost box:
> > ```
> >   CephPoolDefaultPgNum: 32
> >   CephPoolDefaultSize: 1
> > ```
> 32 is indeed better as it allows more PG to be created but I think 8 was enough for my testing...

We can make it 32 when using a dedicated device and 8 otherwise, if it make sense.

EmilienM:
LGTM now, I think it's good to go.

PR #56, opened 2021-05-07
Ceph fixes and enable by default

mdbooth:


EmilienM:
:-1: so for this one, we agreed with @mdbooth that we would disable Nova RBD if we will enable Ceph by default with a loop device backend, because of the performances issues.

mdbooth:
> 👎 so for this one, we agreed with @mdbooth that we would disable Nova RBD if we will enable Ceph by default with a loop device backend, because of the performances issues.

Incidentally, I would put good money on loopback devices having nothing to do with the performance of ceph on a standalone host. The kernel can optimise this to the point that the overhead is inconsequential compared to the overhead of moving from direct device access to a network protocol. Incidentally, I did enable directio on the loopback devices which just avoids an unnecessary additional level of page cache.

The only change I plan here is to add an option to allow a block device to be specified which would replace the loopback devices if given.

EmilienM:
> > -1 so for this one, we agreed with @mdbooth that we would disable Nova RBD if we will enable Ceph by default with a loop device backend, because of the performances issues.
> 
> Incidentally, I would put good money on loopback devices having nothing to do with the performance of ceph on a standalone host. The kernel can optimise this to the point that the overhead is inconsequential compared to the overhead of moving from direct device access to a network protocol. Incidentally, I did enable directio on the loopback devices which just avoids an unnecessary additional level of page cache.

fair enough...

> The only change I plan here is to add an option to allow a block device to be specified which would replace the loopback devices if given.

Most of our baremetal servers have second disks, we wouldn't need a loop device, maybe add an option to not create the loop device?

mdbooth:
@EmilienM @mandre added a patch to configure block devices if you really want that. Is this patch good to go now?

EmilienM:
@mdbooth maybe let's wait for the manila to merge and rebase on top this one.

EmilienM:
I feel like I broke something in this PR

PR #57, opened 2021-05-10
Configure the right MTU for br-ex

EmilienM:
If the interface that is used for br-ex (e.g. eth0) had a specific MTU,
we need to configure the bridge with the right MTU value otherwise we
force it to 1500 and it can causes issues.

### DIFF
--- a/playbooks/templates/dev-install_net_config.yaml.j2
+++ b/playbooks/templates/dev-install_net_config.yaml.j2
@@ -7,7 +7,7 @@ network_config:
 - type: ovs_bridge
   name: br-ex
   use_dhcp: true
-  mtu: 1500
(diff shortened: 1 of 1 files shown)


playbooks/templates/dev-install_net_config.yaml.j2
```diff
@@ -7,7 +7,7 @@ network_config:
 - type: ovs_bridge
   name: br-ex
   use_dhcp: true
-  mtu: 1500
+  mtu: {{ network_info.public_ipv4.mtu }}
```
adduarte, line 10: you don't technically need this, since if you leave it out, it will be derived from the interfaces in the "members" list. which is 
network_info.public_ipv4.interface 
but there is nothin wrong with explicitely setting it I guess.

PR #58, opened 2021-05-11
Migrate a host to CentOS stream && manage container-tools

EmilienM:
For our users on CentOS, we need to ensure that they're on Stream or
they'll have packaging issue during TripleO deployment.

Same for container-tools, it's expected that we use 3.0 from now to get
newest Podman.

### DIFF
--- a/playbooks/prepare_host.yaml
+++ b/playbooks/prepare_host.yaml
@@ -48,6 +48,33 @@
           owner: stack
           group: stack
 
+  # To ensure that our CentOS users are actually using Stream.
+  # TripleO Master only support Stream now.
+  - name: Migrate to CentOS Stream if not done already
+    when:
+      - ansible_facts.distribution == 'CentOS'
+      - ansible_facts.distribution_release != 'Stream'
+    shell: |
+      if ! rpm --query centos-stream-release; then
+          dnf -y swap centos-linux-repos centos-stream-repos
+          dnf -y distro-sync
+      fi
+    args:
+      warn: false
+    register: stream_output
+    changed_when: (stream_output.stdout_lines | length) > 1
+
+  # To get latest version of podman & dependencies we need this
+  # version for now.
+  - name: Ensure that container-tools 3.0 is being used
+    when:
+      - ansible_facts.distribution == 'CentOS'
+    shell: |
+      dnf module disable -y container-tools:rhel8
+      dnf module enable -y container-tools:3.0
+    args:
+      warn: false
+
   - name: Prepare host on CentOS system
     when:
       - ansible_facts.distribution == 'CentOS'


playbooks/prepare_host.yaml
```diff
@@ -48,6 +48,25 @@
           owner: stack
           group: stack
 
+  # To ensure that our CentOS users are actually using Stream.
+  # TripleO Master only support Stream now.
+  - name: Migrate to CentOS Stream if not done already
```
mandre, line 53: We should probably reboot the node for the distro change to fully take effect.

The command is also printing a warning, we should consider setting `warn: false` as the dnf module apparently [doesn't support the two commands used here](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/dnf_module.html).

```
[WARNING]: Consider using the dnf module rather than running 'dnf'.  If you need to use command because dnf is insufficient you can add 'warn: false' to this command task or set 'command_warnings=False' in ansible.cfg to get rid of this message.
```

playbooks/prepare_host.yaml
```diff
@@ -48,6 +48,25 @@
...
+    when:
+      - ansible_facts.distribution == 'CentOS'
+      - ansible_facts.distribution_release != 'Stream'
```
mandre, line 56: Even after a reboot of my now CentOS Stream node, I still get 
```
"ansible_distribution": "CentOS",                                                                                                                               
"ansible_distribution_file_parsed": true,
"ansible_distribution_file_path": "/etc/redhat-release",
"ansible_distribution_file_variety": "RedHat",
"ansible_distribution_major_version": "8",
"ansible_distribution_release": "NA",
"ansible_distribution_version": "8",
```

Perhaps we should parse `/etc/redhat-release` instead?
```
[centos@manila-cloud ~]$ cat /etc/redhat-release 
CentOS Stream release 8
```

EmilienM: yeah that fact was integrated in the latest version of Ansible.
Honestly, I wish we could avoid that PR, I'll mark it as WIP for now.

Once we can get CentOS 8 stream on VEXXHOST, we wouldn't need that.

playbooks/prepare_host.yaml
```diff
@@ -48,6 +48,29 @@
           owner: stack
           group: stack
 
+  # To ensure that our CentOS users are actually using Stream.
+  # TripleO Master only support Stream now.
+  - name: Migrate to CentOS Stream if not done already
+    warn: false
```
EmilienM, line 54: this doesn't work for me

EmilienM:
This one is ready for review, sorry for the mess with git.

mandre (commented):
Thanks for making the changes, looks good.

PR #61, opened 2021-05-13
Add pre-deploy validations

EmilienM:
Add some variable validations, to prevent deployment issues later:

* standalone_host matches with ansible_host
* control_plane_ip is different from public_api
* standalone_role file actually exists

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,7 @@ usage:
 	@echo
 	@echo 'Individual install phase targets:'
 	@echo '  local_requirements: Install Ansible requirements required to run dev-install'
+	@echo '  validations: Run some validations before the deployment'
 	@echo '  prepare_host: Host configuration required before installing standalone, including rhos-release'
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -46,12 +47,16 @@ local-overrides.yaml:
...
 .PHONY: osp_full
-osp_full: local_requirements prepare_host network install_stack prepare_stack
+osp_full: local_requirements validations prepare_host network install_stack prepare_stack
 
 .PHONY: local_requirements
 local_requirements: inventory.yaml local-overrides.yaml
 	$(ANSIBLE_CMD) playbooks/local_requirements.yaml
 
+.PHONY: validations
+validations: inventory.yaml local-overrides.yaml
```
EmilienM, line 57: once https://github.com/shiftstack/dev-install/pull/60 merges, we'll need to change `local-overrides.yaml` to `$(overrides)`.

mdbooth (approved):

PR #63, opened 2021-05-13
Automate manila share type creation

EmilienM:
If Manila is enabled, let's create the share type so the user can start
using manila without creating it manually.


mdbooth (approved):
/lgtm

PR #66, opened 2021-05-14
Fix Ganesha binding when Manila is enabled

EmilienM:
We need to bind Ganesha to the control plane or the NFS mounts are not
binding on the control plane, and then the client will try to use the
public IP to mount, and it's wrong.

Tested against a real environment.


playbooks/install_stack.yaml
```diff
@@ -84,10 +84,10 @@
     # This works round https://bugs.launchpad.net/tripleo/+bug/1911022
```
EmilienM, line 84: I'll have to remove that comment probably

mandre (approved):

PR #67, opened 2021-05-14
Download CIP from upstream/master if cip_config not defined on CentOS

EmilienM:
We already have safe defaults on RHEL/OSP for cip_config, where if it's
not defined we download the CIP from internal puddles and it's fine.

However when deploying on CentOS, we don't have safe defaults, so let's
add safe defaults which are:

* using CIP from tripleo-common master
* tripleo_repos_stream set to True


EmilienM:
let me test it first

EmilienM:
tested, it works

mandre:
Thanks.

PR #70, opened 2021-05-17
Allow to override openshift password in Keystone

EmilienM:
For security purposes, it can be useful to override the default
password "password" that was set before with something more secure.


mandre (approved):
:tada:

PR #75, opened 2021-05-18
Enable TripleO HA services by default

EmilienM:
In OSP16, HAproxy isn't deployed with Pacemaker by default which
apparently can cause issues if the public IP is the same as the control
plane IP.

To counter that, we simply need to turn the services that can be HA to
be deployed by Pacemaker.

Also, it removes the need of having the control plan IP to be the same
as the public IP.


EmilienM:
tested against OSP 16.2.

mdbooth (approved):
/lgtm

PR #76, opened 2021-05-18
WIP - DO NOT MERGE - Use Ansible async when migrating to CentOS Stream

EmilienM:
It has been observed that when running dev-install from Github action
workflows, the task that update the host to CentOS never finishes and we
think that it's because the task takes too much time so the connection
is somehow lost.

Using async lets us follow the progress of the task and poll every
3 seconds what the status is.


EmilienM:
not needed for now, might re-open.

PR #78, opened 2021-05-19
Enable SSL support

EmilienM:
By setting `ssl_enabled` to true, dev-install will do the following:

* Generate a CA, re-using what Matt did when doing `make certs`.
* Generate a private key, a certificate and sign with the CA.
* The certificate is signed to allow DNS and IP addresses, for
  control plane IP, public IP and the standalone_host, so the
  certificate will be usable for different kind of endpoints
  even though OpenStack TLS will be configured only with IP based
  endpoints for Public interface only.
* When running `make local_os_client`, it will pull the CA from the host
  on the local host so the client can use the OpenStack out of the box
  using SSL.

  This was manually tested with OSP 16.2.

### DIFF
--- a/playbooks/install_stack.yaml
+++ b/playbooks/install_stack.yaml
@@ -28,6 +28,62 @@
         ha_env:
         - /usr/share/openstack-tripleo-heat-templates/environments/docker-ha.yaml
 
+  - name: Enable SSL
+    when: ssl_enabled
+    block:
+    - name: Add SSL service
+      set_fact:
+        service_envs: "{{ service_envs | union(ssl_env) }}"
+      vars:
+        ssl_env:
+        - /usr/share/openstack-tripleo-heat-templates/environments/ssl/tls-endpoints-public-ip.yaml
+        - /usr/share/openstack-tripleo-heat-templates/environments/ssl/enable-tls.yaml
+        - /usr/share/openstack-tripleo-heat-templates/environments/ssl/inject-trust-anchor.yaml
+
+  - name: Generate SSL self-signed certificate
+    when: ssl_enabled
+    include_role:
+      name: simpleca
+    vars:
+      cert_user: standalone
+      ca_dir: "{{ ansible_env.HOME }}/ssl/ca"
+      cert_dir: "{{ ansible_env.HOME }}/ssl"
+      cert_name: standalone
+
+  - name: Prepare the host for SSL
+    when: ssl_enabled
+    block:
+      - name: Read SSL certificate
+        slurp:
+          src: "{{ ansible_env.HOME }}/ssl/standalone.crt"
+        register: ssl_cert_output
+      - name: Read SSL key
+        slurp:
+          src: "{{ ansible_env.HOME }}/ssl/standalone.key"
+        register: ssl_key_output
+      - name: Read CA certificate
+        slurp:
+          src: "{{ ansible_env.HOME }}/ssl/ca/simpleca.crt"
+        register: ssl_ca_cert_output
+      - name: Set facts for SSL
+        set_fact:
+          ssl_cert: "{{ ssl_cert_output['content'] | b64decode }}"
+          ssl_key: "{{ ssl_key_output['content'] | b64decode }}"
+          ssl_ca_cert: "{{ ssl_ca_cert_output['content'] | b64decode }}"
+      - name: Copy CA cert into PKI
+        become: true
+        become_user: root
+        copy:
+          dest: "{{ ssl_ca_cert_path }}"
+          content: "{{ ssl_ca_cert }}"
+          mode: '0444'
+          owner: root
+          group: root
+      - name: Update CA trust
+        become: true
+        become_user: root
+        command: update-ca-trust extract
+
   - name: Install the tripleo client
     yum:
       name:
@@ -227,6 +283,8 @@
       tripleo_deploy_output_dir: "{{ ansible_env.HOME }}"
       tripleo_deploy_local_ip: "{{ public_api }}"
       tripleo_deploy_control_virtual_ip: "{{ control_plane_ip }}"
+      #TODO(emilien) figure out if we need this:
+      #tripleo_deploy_public_virtual_ip: "{{ ssl_enabled | ternary(public_api, omit) }}"
       tripleo_deploy_environment_files: "{{ default_tripleo_envs + service_envs + tripleo_override_envs }}"
       tripleo_deploy_generate_scripts: true
       tripleo_deploy_keep_running: true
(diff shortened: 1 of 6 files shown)


playbooks/roles/simpleca/tasks/main.yaml
```diff
@@ -38,7 +38,11 @@
     path: "{{ cert_dir }}/{{ cert_user }}.csr"
     privatekey_path: "{{ user_key.filename }}"
     common_name: "{{ cert_name }}"
-    subject_alt_name: "DNS:{{ cert_name }}"
+    subject_alt_name:
+      - "DNS:{{ cert_name }}"
+      - "DNS:{{ standalone_host }}"
+      - "IP:{{ public_api }}"
+      - "IP:{{ control_plane_ip }}"
```
EmilienM, line 45: in reality I think we only need this one ^
but added the other ones in case we need to re-use the certificate. It's for dev anyway?

playbooks/vars/defaults.yaml
```diff
@@ -94,6 +94,11 @@ octavia_enabled: true
 
 manila_enabled: false
 
+# Enable SSL for the OpenStack public endpoints
+# using self-signed certificate
+ssl_enabled: false
```
EmilienM, line 99: I think we should enable this by default but first I want people to test it and see if it breaks for anyone. So please don't ask me to enable it by default now, and let's wait a bit.

playbooks/install_stack.yaml
```diff
@@ -28,6 +28,62 @@
...
+      vars:
+        ssl_env:
+        - /usr/share/openstack-tripleo-heat-templates/environments/ssl/tls-endpoints-public-ip.yaml
+        - /usr/share/openstack-tripleo-heat-templates/environments/ssl/enable-tls.yaml
+        - /usr/share/openstack-tripleo-heat-templates/environments/ssl/inject-trust-anchor.yaml
+
+  - name: Generate SSL self-signed certificate
+    when: ssl_enabled
+    include_role:
+      name: simpleca
```
EmilienM, line 46: thank you Matt I was able to re-use your stuff :)

playbooks/install_stack.yaml
```diff
@@ -227,6 +283,8 @@
       tripleo_deploy_output_dir: "{{ ansible_env.HOME }}"
       tripleo_deploy_local_ip: "{{ public_api }}"
       tripleo_deploy_control_virtual_ip: "{{ control_plane_ip }}"
+      #TODO(emilien) figure out if we need this:
```
mandre, line 286: So, is it needed or not?

playbooks/local_os_client.yaml
```diff
@@ -16,6 +16,33 @@
...
+    when: ssl_enabled
+    block:
+      - name: Read CA certificate
+        slurp:
+          src: "{{ ssl_ca_cert_path }}"
+        register: ssl_ca_cert_output
+      - name: Set fact for CA cert
+        set_fact:
+          ssl_ca_cert: "{{ ssl_ca_cert_output['content'] | b64decode }}"
+      - name: Copy CA cert into PKI
```
mandre, line 29: Hmmm, can we not do that, or at least make it optional? We don't need the certificate in the host trust bundle starting from OCP 4.7 and this is probably something we want to test. I also don't want to pollute my local trust bundle.

We should instead add the path to the CA Cert  in the generated `clouds.yaml` entry with the `cacert` key.

playbooks/templates/standalone_parameters.yaml.j2
```diff
@@ -80,6 +80,17 @@ parameter_defaults:
...
 {% endif %}
+{% if ssl_enabled %}
+  AddVipsToEtcHosts: true
+  HorizonSecureCookies: True
+  SSLCertificate: |
+    {{ ssl_cert | indent }}
+  SSLKey: |
+    {{ ssl_key | indent }}
+  SSLRootCertificate: |
+    {{ ssl_ca_cert | indent }}
```
mandre, line 91: Possibly these things live as a secret in `ci-configs` and we pass them as parameters to ansible.
In other words, we may need to allow passing an existing cert when deploying the OpenStack platform for our CI needs since it could be difficult to dynamically update the cert referenced by the `clouds.yaml` file in CI. To investigate.

playbooks/local_os_client.yaml
mdbooth: I agree we should put this in clouds.yaml by preference. However, I suspect we'll need it in the trust bundle for some use cases anyway. e.g. Do we have openstack cli support for manila?

In short: if this were the only thing holding this patch up I'd merge it.

mdbooth (approved):
/lgtm

Not merging for Martin's queries.

playbooks/install_stack.yaml
EmilienM: so it doesn't do do anything, the public endpoint still is on the control plane IP... I'll investigate later

playbooks/local_os_client.yaml
EmilienM: @mandre FYI the `cacert` entry in clouds.yaml is already added when SSL is [configured in TripleO](https://opendev.org/openstack/tripleo-heat-templates/src/branch/master/deployment/keystone/keystone-container-puppet.yaml#L777-L781); so we don't need to take care of it.

So if I understand the suggestion, the ask is to make this step optional; so we can easily enable it but disable it by default. If that's the case, I agree and I'll do it.

playbooks/templates/standalone_parameters.yaml.j2
EmilienM: ack, I'll add support for this.

playbooks/local_os_client.yaml
mdbooth: We'll presumably need to rewrite the cacert path in clouds.yaml for both the standalone and standalone_openshift users.

EmilienM: > We'll presumably need to rewrite the cacert path in clouds.yaml for both the standalone and standalone_openshift users.

why that?

EmilienM: the content is already good, I checked when testing. I didn't have to change anything manually.

mandre: We should probably store the cert to a local directory owned by the user rather than requiring password-less sudo. In that case we would need to update the cacert path as Matt noted.

EmilienM: I honestly have mixed feelings about this request, just because if the user finally decides to update their PKI with `update-ca-trust extract`, the CA cert needs to be in /etc/pki, not somewhere in $HOME.

Maybe we can improve this later?

mandre (approved):
LGTM

PR #80, opened 2021-05-20
Use current-tripleo by default

EmilienM:
Instead of current-tripleo-dev which is actual master, let's use
current-tripleo which is a repo that is tested, from master, so might
be lagging a bit but at least better tested.


EmilienM:
this is tested and OK to go

mandre (approved):
I had an issue this morning where current-tripleo-dev was broken [1] so I fully support we stop following head and use a tested build.

[1] https://bugs.launchpad.net/tripleo/+bug/1928968

PR #82, opened 2021-05-20
Support TripleO deployment from stable/train

EmilienM:
2 changes:

### Update example overrides
To help anyone who wants to deploy TripleO train, will be able to use the example.

### ContainerCli
When deploying TripleO upstream from stable/train, we need to override
the default ContainerCli to "podman" and not let "docker" since it's not
a thing anymore.

We kept it this way for backward compatibility in TripleO but in reality
when enabling HA on centos8 in Train, you need Podman.


mdbooth (approved):
/lgtm

PR #84, opened 2021-05-22
Support for RHSM

EmilienM:
When we want to deploy Red Hat OpenStack as customers would do, using
the product, we don't want to use the internal puddle (rhos-release).
For that we need to use Red Hat Subscription Manager (RHSM) which has
its Ansible role.

This patch will:

* deploy the role in ~/.ansible/roles
* provide safe defaults for OSP 16.1 (latest release)
* call the role when rhsm_enabled is set to True
* manage required modules
* generate CIP config using CLI
* login to the containe registry when deploying

More information about the role:
https://docs.openstack.org/project-deploy-guide/tripleo-docs/latest/features/rhsm.html

### DIFF
--- a/playbooks/install_stack.yaml
+++ b/playbooks/install_stack.yaml
@@ -227,11 +227,19 @@
           octavia_env:
             - /usr/share/openstack-tripleo-heat-templates/environments/services/octavia.yaml
 
+  - name: Generate container_image_prepare.yaml if not using rhos-release
+    when:
+      - cip_config is not defined
+      - ansible_facts.distribution == 'RedHat'
+      - rhsm_enabled
+    command: openstack tripleo container image prepare default --output-env-file /home/stack/containers-prepare-parameters.yaml
+
   # On RHEL/OSP if cip_config is not provided, we download it from latest puddle
   - name: Set cip_config from downloaded container_image_prepare.yaml
     when:
       - cip_config is not defined
       - ansible_facts.distribution == 'RedHat'
+      - not rhsm_enabled
     block:
     - name: Read container_image_prepare.yaml
       slurp:
--- a/playbooks/prepare_host.yaml
+++ b/playbooks/prepare_host.yaml
@@ -20,9 +20,27 @@
       key: "{{ item }}"
     loop: "{{ authorized_keys }}"
 
-  - name: Prepare host on RHEL system
+  - name: Prepare host on RHEL system with Subscription Manager
     when:
       - ansible_facts.distribution == 'RedHat'
+      - rhsm_enabled
+    block:
+      - name: Configure Red Hat Subscription Manager
+        import_role:
+          name: redhat-subscription
+      - name: Install container-tools module
+        shell: |
+          dnf module disable -y container-tools:rhel8
+          dnf module enable -y container-tools:"{{ rhsm_container_tools_version }}"
+      - name: Install virt module
+        shell: |
+          dnf module disable -y virt:rhel
+          dnf module enable -y virt:"{{ rhsm_release }}"
+
+  - name: Prepare host on RHEL system with rhos-release
+    when:
+      - ansible_facts.distribution == 'RedHat'
+      - not rhsm_enabled
     block:
       - name: install rhos-release
         yum:
--- a/playbooks/templates/standalone_parameters.yaml.j2
+++ b/playbooks/templates/standalone_parameters.yaml.j2
@@ -81,6 +81,12 @@ parameter_defaults:
   CephPoolDefaultPgNum: 8
   CephPoolDefaultSize: 1
 {% endif %}
+{% if rhsm_enabled %}
+  ContainerImageRegistryCredentials:
+    registry.redhat.io:
+      {{ redhat_registry_credentials | mandatory }}
+  ContainerImageRegistryLogin: true
+{% endif %}
 {% if ssl_enabled %}
   AddVipsToEtcHosts: true
   HorizonSecureCookies: True
--- a/playbooks/local_requirements.yaml
+++ b/playbooks/local_requirements.yaml
@@ -2,5 +2,7 @@
   connection: local
   vars:
     update_operator: true
+    update_rhsm: true
   roles:
     - operators
+    - rhsm
--- a/playbooks/roles/rhsm/tasks/main.yaml
+++ b/playbooks/roles/rhsm/tasks/main.yaml
@@ -1,0 +1,16 @@
+---
+- name: Ensure we have ~/.ansible
+  tags:
+    - always
+    - lab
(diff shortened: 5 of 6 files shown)


playbooks/templates/standalone_parameters.yaml.j2
```diff
@@ -81,6 +81,12 @@ parameter_defaults:
   CephPoolDefaultPgNum: 8
   CephPoolDefaultSize: 1
 {% endif %}
+{% if rhsm_enabled %}
+  ContainerImageRegistryCredentials:
+    registry.redhat.io:
+      {{ rhsm_username }}: {{ rhsm_password }}
```
mandre, line 87: These two variables might be uninitialized.

EmilienM: I made them required for now.

EmilienM:
PR tested manually in PSI, using my employee SKU credentials.

playbooks/templates/standalone_parameters.yaml.j2
```diff
@@ -81,6 +81,12 @@ parameter_defaults:
   CephPoolDefaultPgNum: 8
   CephPoolDefaultSize: 1
 {% endif %}
+{% if rhsm_enabled %}
+  ContainerImageRegistryCredentials:
+    registry.redhat.io:
+      {{ rhsm_username | mandatory }}: "{{ rhsm_password | mandatory }}"
```
mdbooth, line 87: This is going to require people to put rhsm_password in a plain text file. I don't think we can do this. I know other RH services require similar, but we shouldn't add to the mess.

What other ways are there to pass these credentials? Reading from a file descriptor when the process is launched would be ideal, but even something like an environment variable would be better than persisting it in plain text.

mdbooth (changes requested):
I like this, but we need to...

/hold for the password issue

playbooks/templates/standalone_parameters.yaml.j2
EmilienM: This is the way it's implemented in TripleO, I'm not sure how you would do otherwise unless we use something like ansible-vault or something.

mdbooth: I was actually just thinking about local-overrides.yaml, but of course this is also going to write it into the template.

That's really bad design. Can we generate a pull secret that contains single-purpose credentials here? Or is plain text authentication the only supported method? Plain text passwords doesn't sound like something that would fly with many customers, so I'm convinced there's another way to do this.

EmilienM: Our customers are writing Heat environments with RHSM credentials since OSP 7 I believe. I'm not sure what to say. Yes this is bad design but I don't think we can solve it here.
I might have another alternative proposal where we don't use ContainerImageRegistryCredentials to login against the container registries, but manually run "podman login" during the deployment (using Ansible in dev-install), so we only "store" credentials in dev-install overrides (which can be encrypted via git-crypt) and not in clear in the Heat environments on the host where OpenStack is deployed.

If we go that route, we'll have to test it (I think it'll work but not 100% sure).

EmilienM: I updated it to have a new parameter `redhat_registry_credentials` that will be used exclusively for RH container registry credentials.

playbooks/install_stack.yaml
```diff
@@ -205,11 +205,19 @@
           octavia_env:
             - /usr/share/openstack-tripleo-heat-templates/environments/services/octavia.yaml
 
+  - name: Generate container_image_prepare.yaml if not using rhos-release
+    when:
+      - cip_config is not defined
+      - ansible_facts.distribution == 'RedHat'
+      - rhsm_enabled
+    command: openstack tripleo container image prepare default --output-env-file /home/stack/containers-prepare-parameters.yaml
```
mandre, line 213: I've seen `/home/stack` being hardcoded in a few places and `{{ ansible_env.HOME }}` at other places. I think we should prefer `{{ ansible_env.HOME }}` for consistency:
```suggestion
    command: openstack tripleo container image prepare default --output-env-file {{ ansible_env.HOME }}/containers-prepare-parameters.yaml
```

No need to make the change here, we should make the change globally in a separate PR.

playbooks/prepare_host.yaml
```diff
@@ -20,9 +20,27 @@
...
-  - name: Prepare host on RHEL system
+  - name: Prepare host on RHEL system with Subscription Manager
     when:
       - ansible_facts.distribution == 'RedHat'
+      - rhsm_enabled
+    block:
+      - name: Configure Red Hat Subscription Manager
+        import_role:
+          name: redhat-subscription
+      - name: Install container-tools virt module
```
mandre, line 31: Nit
```suggestion
      - name: Install container-tools module
```

mandre (commented):
LGTM.

@mdbooth I believe @EmilienM addressed your concerns for the rhsm credentials, can you check?

playbooks/prepare_host.yaml
```diff
@@ -20,9 +20,27 @@
...
 
-  - name: Prepare host on RHEL system
+  - name: Prepare host on RHEL system with Subscription Manager
     when:
       - ansible_facts.distribution == 'RedHat'
+      - rhsm_enabled
+    block:
+      - name: Configure Red Hat Subscription Manager
+        import_role:
+          name: redhat-subscription
```
mdbooth, line 30: I can't work out how this would authenticate to RHSM. We're talking about including subscription manager. Where would that go?

Or for the moment are we saying that you'd have to log in to RHSM manually on the host before executing the ansible? I'd be ok with that until we sort activation keys out, btw.

playbooks/install_stack.yaml
EmilienM: I'll do it.

playbooks/prepare_host.yaml
EmilienM: This role will execute RHSM for you in the first thing, so later we can use DNF to pull rpms. I tested and it works.
As for activation keys, the user can define the Ansible vars for that role to use activation keys so no work will be needed to make it work with it.

mdbooth: 👆 is my remaining concern here.

I just asked @EmilienM how this works. The redhat-subscription role implicitly uses vars which must be defined in local-overrides. This isn't ideal (we should get activation keys working), but as these credentials are never persisted anywhere except the user's own workstation we could let it slide briefly for testing purposes. However, we *MUST* replace this with activation keys asap, and we *MUST NOT* integrate anything into CI which uses user credentials.

mdbooth (approved):

PR #86, opened 2021-05-25
Make ceph vg creation idempotent

mdbooth:
* Split out creation task into multiple tasks
* Use direct io on loopback device
* Only ever create loopback device with systemd unit
* Use ansible modules for vg/lv creation
* Rename created volume group to vg_ceph

### DIFF
--- a/playbooks/prepare_host.yaml
+++ b/playbooks/prepare_host.yaml
@@ -130,14 +130,10 @@
           - util-linux
           - lvm2
         state: present
-    - name: Use dd and losetup to create the loop devices
-      shell: |
-        fallocate -l {{ ceph_loop_device_size }}G /var/lib/ceph-osd.img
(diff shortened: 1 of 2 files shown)


playbooks/prepare_host.yaml
```diff
@@ -148,13 +144,29 @@
...
+    - name: Create ceph data LV
+      lvol:
+        vg: vg_ceph
+        lv: data
+        size: 96%VG
+    - name: Create ceph db LV
+      lvol:
+        vg: vg_ceph
+        lv: db
+        size: 4%VG
```
EmilienM, line 172: this won't work, I tested it when implementing this thing and you need to keep space for some metadatas I think. Which is why I initially did `(ceph_loop_device_size * 4 / 100) - 0.1`

mdbooth: It does work. I'm using it.

PR #87, opened 2021-05-25
Create and use user-local copy of ca cert

mdbooth:
This allows referencing the CA cert without putting it in the local host
trust bundle.

/cc @mandre

### DIFF
--- a/playbooks/local_os_client.yaml
+++ b/playbooks/local_os_client.yaml
@@ -74,18 +74,41 @@
         set_fact:
           cloudsyaml: "{{ {'clouds': {}} }}"
 
+  - name: Configure cacert locally
+    when: ssl_enabled
+    block:
+    - name: Set local path of cacert
+      set_fact:
+        cacert_path: "{{ ansible_env.HOME }}/.config/openstack/{{ local_cloudname }}-ca.crt"
+
+    - name: Copy CA cert into local config directory
+      copy:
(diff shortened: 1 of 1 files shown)


playbooks/local_os_client.yaml
```diff
@@ -74,18 +74,41 @@
...
+    when: ssl_enabled
+    block:
+    - name: Set local path of cacert
+      set_fact:
+        cacert_path: "{{ ansible_env.HOME }}/.config/openstack/{{ local_cloudname }}-ca.crt"
+
+    - name: Copy CA cert into local config directory
+      copy:
+        dest: "{{ cacert_path }}"
+        content: "{{ hostvars['standalone']['ssl_ca_cert'] }}"
```
EmilienM, line 87: `ssl_ca_cert` is defined in `playbooks/install_stack.yaml` playbook, I don't think it'll actually work here (you should get an error saying this variable doesn't exist).

EmilienM: if we're doing that, do we want to keep `Copy CA cert into PKI` task?

mdbooth: We slurped it earlier in this playbook on line 28.

EmilienM:
If you set `update_local_pki` to False (the default), the CA won't be in the local host trust bundle. So why do we need this PR?

PR #89, opened 2021-05-25
Disable interfaces which aren't in use

mdbooth:
I can't install without this patch. deploy always fails with TripleO complaining that the network service failed.

Bonus: this patch auto-detects SR-IOV devices. We can use that to improve SR-IOV support.

### DIFF
--- a/playbooks/roles/network_info/tasks/main.yaml
+++ b/playbooks/roles/network_info/tasks/main.yaml
@@ -34,3 +34,59 @@
       content: "{{ network_info | to_nice_json }}"
       mode: 0755
   when: networkinfo_file.failed
+
+- name: Find network devices with a device link
+  find:
+    paths: /sys/class/net
+    patterns:
+    - device
+    file_type: link
+    follow: true
+    depth: 2
+    recurse: true
+    use_regex: false
+  register: device_links
+
+- name: Extract paths from device_links
+  set_fact:
+    device_links: "{{ device_links.files | map(attribute='path') | list }}"
+
+- name: Extract device names from device_links
+  set_fact:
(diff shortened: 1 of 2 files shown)


playbooks/roles/network_info/tasks/main.yaml
```diff
@@ -34,3 +34,66 @@
...
+  register: sriov_physfndir
+
+- name: Extract device names for SR-IOV VFs
+  set_fact:
+    sriov_vfs: "{{ sriov_vfs + [ item | dirname | dirname | basename ] }}"
+  loop: "{{ sriov_physfndir.files | map(attribute='path') | list }}"
+
+- name: Get a list of physical network device names
+  set_fact:
+    physical_nics: "{{ physical_nics + [(item | dirname | basename)] }}"
```
EmilienM, line 96: where do we use this fact?

playbooks/roles/network_info/tasks/main.yaml
```diff
@@ -34,3 +34,66 @@
...
+  find:
+    paths: "{{ device_links }}"
+    patterns:
+    - physfn
+    file_type: link
+  register: sriov_physfndir
+
+- name: Extract device names for SR-IOV VFs
+  set_fact:
+    sriov_vfs: "{{ sriov_vfs + [ item | dirname | dirname | basename ] }}"
```
EmilienM, line 91: where do we use this fact?

mdbooth: In the next commit.

playbooks/roles/network_info/tasks/main.yaml
mdbooth: Whoops, looks like a duplicate. I removed it.

PR #90, opened 2021-05-25
Make m1.medium usable as a worker node

mdbooth:


playbooks/prepare_stack.yaml
```diff
@@ -64,7 +64,7 @@
           openstack flavor create --ram 1024 --disk 10 --vcpu 2 --public m1.tiny
       fi
       if ! openstack flavor show m1.medium; then
-          openstack flavor create --ram 4096 --disk 20 --vcpu 2 --public m1.medium
+          openstack flavor create --ram 4096 --disk 25 --vcpu 2 --public m1.medium
```
EmilienM, line 67: I don't think you can deploy a worker with 4GB of RAM. Docs says 8GB minimum.

mdbooth: Ack. I'll drop this PR.

EmilienM: For reference:
https://docs.openshift.com/container-platform/4.7/installing/installing_openstack/installing-openstack-installer-custom.html#prerequisites
```
A flavor with at least 16 GB memory, 4 vCPUs, and 25 GB storage space
```

PR #92, opened 2021-05-26
SSL: "Bring your own CA" support & generate SSL locally

EmilienM:
# Bring your own CA
Allow a user to bring their own Certificate Authority to dev-install,
so when OpenStack will be deployed, the generated certificate will be
signed with this CA.

It's a famous customer use-case, and will also be used by our CI, so we
can rely on a stable CA and just generate ephemeral certificates when
deploying OpenStack.

# Generate SSL files locally instead of remotely

Instead of generating all SSL files remotely, do it locally, it's more
secure.

The main reason of doing this is because we do not want the CA private
key to be on the remote host, this would be a security issue in case
someone has access to the host, they can compromise our CA

### DIFF
--- a/playbooks/install_stack.yaml
+++ b/playbooks/install_stack.yaml
@@ -40,35 +40,57 @@
         - /usr/share/openstack-tripleo-heat-templates/environments/ssl/enable-tls.yaml
         - /usr/share/openstack-tripleo-heat-templates/environments/ssl/inject-trust-anchor.yaml
 
-  - name: Generate SSL self-signed certificate
+  - name: Generate SSL self-signed certificate on localhost
     when: ssl_enabled
-    include_role:
-      name: simpleca
-    vars:
-      cert_user: standalone
-      ca_dir: "{{ ansible_env.HOME }}/ssl/ca"
-      cert_dir: "{{ ansible_env.HOME }}/ssl"
-      cert_name: standalone
+    become: false
+    # We run this block on localhost because we don't want to put the CA key on the remote
+    # server, which could lead to security problems.
+    delegate_to: localhost
+    block:
+    - name: Create temporary directory for SSL files
+      tempfile:
+        state: directory
+        suffix: dev-install-ssl
+      register: ssl_dir
+    - name: Generate SSL self-signed certificate
+      include_role:
+        name: simpleca
+      vars:
+        cert_user: standalone
+        ca_dir: "{{ ssl_dir.path }}/ssl/ca"
+        cert_dir: "{{ ssl_dir.path }}/ssl"
+        cert_name: standalone
 
   - name: Prepare the host for SSL
     when: ssl_enabled
     no_log: true
     block:
-      - name: Read SSL certificate
-        slurp:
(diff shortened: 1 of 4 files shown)


EmilienM:
I'm still testing it now and I think I found an issue, please don't merge it yet.

EmilienM:
tested in PSI and it worked fine.

playbooks/install_stack.yaml
```diff
@@ -40,35 +40,58 @@
...
+    # server, which could lead to security problems.
+    delegate_to: localhost
+    block:
+    - name: Create temporary directory for SSL files
+      tempfile:
+        state: directory
+        suffix: dev-install-ssl
+      register: ssl_dir
+    - name: Generate SSL self-signed certificate
+      when: ssl_enabled
```
mandre, line 56: No need, the parent task already has `when: ssl_enabled`

playbooks/install_stack.yaml
```diff
@@ -40,35 +40,58 @@
...
+          - name: Read SSL certificate
+            slurp:
+              src: "{{ ssl_dir.path }}/ssl/standalone.crt"
+            register: ssl_cert_output
+            when: ssl_cert is not defined
+          - name: Read SSL key
+            slurp:
+              src: "{{ ssl_dir.path }}/ssl/standalone.key"
+            register: ssl_key_output
+            when: ssl_ssl_key is not defined
```
mandre, line 82: Should be `ssl_key`, I'll address the issue in a separate patch.

mandre (commented):
Happy to merge this once you had a chance to look at my comments.

playbooks/install_stack.yaml
mandre: I meant to create a PR and accidentally pushed to main, sorry :angel: 

https://github.com/shiftstack/dev-install/commit/6724a1daa3b7834871642781913a66672b92d6b7

playbooks/vars/defaults.yaml
```diff
@@ -98,9 +98,13 @@ manila_enabled: false
 ssl_enabled: false
 # To use your own certificates and key, set these variables, otherwise
 # dev-install will generate self-signed certificates and use them
-# ssl_ca_cert
 # ssl_cert
 # ssl_key
+# If you already have your own CA, you can provide these two variables
+# and dev-install will sign the certificates with it.
+# If left unset, dev-install will generate the CA ans self-sign it.
```
mdbooth, line 105: nit: /and/

playbooks/install_stack.yaml
mdbooth: Ah, that will also be why there's a merge conflict.

mdbooth (commented):
Need to fix the merge conflict, obviously. Might as well fix the typo while you're at it, but that's not important.

Something to think about: by generating a new CA every time this executes (when you're not passing a pre-created one in), I think this means install_stack can never be idempotent. It's not idempotent right now and I doubt it ever will be, so I suspect that's not a problem in practise.

mandre (approved):
:+1:

PR #93, opened 2021-05-26
Revert "Add pre-deploy validations"

EmilienM:
This doesn't work well in many cases, let's revert and rework it later.
This reverts commit 110240d8a91d561089165841f03d8a795addcda0.


EmilienM:
@mdbooth it's blocking CI deployment I'm merging it.

PR #96, opened 2021-05-27
UX improvements for security

EmilienM:
## Add local_os_client to osp_full
When installing a cloud, it makes sense to run `local_os_client` by
default so the user can operate the OpenStack cloud immediately.

## Enable SSL by default
To improve the user experience, we enable SSL by default, with
self-signed CA & certificates, so the public endpoints are secure by
default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ $(overrides):
 #
 
 .PHONY: osp_full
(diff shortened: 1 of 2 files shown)


mdbooth (approved):
LGTM

PR #97, opened 2021-05-31
Separate control plane & public endpoints networks

EmilienM:
It is more secure to keep internal/admin networks under a secured
control plane network, and use the public interface for the public
endpoints.

* We will create external-network-vip.yaml environment, until the
  upstream patch in THT will merge:
  https://review.opendev.org/c/openstack/tripleo-heat-templates/+/793836
  This environments helps to configure TripleO to use an external IP for
  the public VIP.

* Configure standalone to run with local_ip, control plane IP and public
  IP, and also load the network config that is done for Undercloud, this
  is fine since TripleO Undercloud already separates public & internal
  endpoints.

* Use 192.168.24.0/24 as default control plane network, keep first IP
  for the local IP and second IP for the control plane VIP.

PR #98, opened 2021-06-02
Allow to plug physical NICs to br-hostonly & br-ctlplane

EmilienM:
It can be useful to plug physical NICs to both bridges (control plane
and provider network), so they can communicate with other nodes.

Use-cases:
* for control plane, in the case of DCN the AZ nodes need to reach the
  central node control plane, so using a physical NIC will be
  necessary.
* for provider network, the workloads plugged on this network might
  want to reach another workload running on another hypervisor,
  therefore it makes sense to plug the provider networks together via
  a physical NIC.


EmilienM:
I'll rebase on top of #97

EmilienM:
@mdbooth @mandre what do you think about this one?

mdbooth:
Rather than bridging these private networks to an external network, could we instead add tunnels to them, and leave the routing of the tunnel traffic to the host? I'm just throwing this out there because it seems weird to bridge these networks: I have no idea if this is possible or practical. I agree we need a solution to this, though.

Perhaps @adduarte would have insight here.

EmilienM:
> Rather than bridging these private networks to an external network, could we instead add tunnels to them, and leave the routing of the tunnel traffic to the host? I'm just throwing this out there because it seems weird to bridge these networks: I have no idea if this is possible or practical. I agree we need a solution to this, though.

Another solution would be to connect OVS Bridges with patch ports, so no tunnels are involved. I'll test it, but honestly I don't think we should spend too much time on that business.

EmilienM:
Or another solution again, is to configure the IP addresses on the dummy interfaces, and put the dummy interfaces under the br-ex bridge. So we don't have a bridge for control plane & provider network.

mdbooth:
I just had a quick look at the OVS docs, and tunnels look pretty simple tbh. Almost certainly simpler than anything we'd hack up. Even an ipsec tunnel looks simple.

EmilienM:
Whatever solution we pick I think it must be integrated with os-net-config, so it's persistent after a reboot and integrated with network scripts on the host.

mdbooth (approved):
I love how simple this is.

EmilienM:
this is now tested.

EmilienM:
```
[root@standalone ~]# ovs-vsctl show
1183f4be-90c6-4425-a794-dacb2910971d
    Bridge br-ctlplane
        fail_mode: standalone
        Port tun-ctlplane-9eb947e6-a1f1-5058-be79-d18ea53077b8
            Interface tun-ctlplane-9eb947e6-a1f1-5058-be79-d18ea53077b8
                type: vxlan
                options: {key="100", remote_ip="192.168.68.132"}
        Port dummy0
            Interface dummy0
        Port br-ctlplane
            Interface br-ctlplane
                type: internal
    Bridge br-hostonly
        fail_mode: standalone
        Port dummy1
            Interface dummy1
        Port tun-hostonly-9eb947e6-a1f1-5058-be79-d18ea53077b8
            Interface tun-hostonly-9eb947e6-a1f1-5058-be79-d18ea53077b8
                type: vxlan
                options: {key="101", remote_ip="192.168.68.132"}
        Port br-hostonly
            Interface br-hostonly
                type: internal
    Bridge br-ex
        fail_mode: standalone
        Port br-ex
            Interface br-ex
                type: internal
        Port eth0
            Interface eth0
    ovs_version: "2.13.4"
```

ping works with the other host.

playbooks/templates/dev-install_net_config.yaml.j2
```diff
@@ -26,6 +26,14 @@ network_config:
   - type: interface
     name: dummy0
     nm_controlled: true
+{% for ip in tunnel_remote_ips | default([]) %}
```
mdbooth, line 29: nit: the default here is redundant, as we already default it in vars/defaults.yaml

playbooks/templates/dev-install_net_config.yaml.j2
```diff
@@ -38,6 +46,14 @@ network_config:
   - destination: {{ hostonly_cidr }}
     nexthop: {{ hostonly_gateway }}
   members:
+{% for ip in tunnel_remote_ips | default([]) %}
```
mdbooth, line 49: ditto

playbooks/templates/dev-install_net_config.yaml.j2
```diff
@@ -26,6 +26,14 @@ network_config:
...
+  - type: ovs_tunnel
+    name: "tun-ctlplane-{{ ip | to_uuid }}"
+    tunnel_type: vxlan
+    ovs_options:
+      - remote_ip={{ ip }}
+      - key=100
+{% endfor %}
 - type: ovs_bridge
   name: br-hostonly
```
mdbooth, line 38: We probably ought to change the name of this bridge eventually.

mdbooth (approved):

playbooks/templates/dev-install_net_config.yaml.j2
EmilienM: ack

playbooks/templates/dev-install_net_config.yaml.j2
EmilienM: ack

PR #100, opened 2021-06-03
added the ability to save static routes preserving the default route

gprocunier:
new task to save nmstate_routes
updated the network_state variable to include static route information

### DIFF
--- a/playbooks/network.yaml
+++ b/playbooks/network.yaml
@@ -50,6 +50,17 @@
         ipv6:
           enabled: false
 
+  # this saves the static route configuration including the default route
+  # prior to filtering later
+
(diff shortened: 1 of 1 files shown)


EmilienM:
I tested it, it works for me. @mdbooth any thoughts?

EmilienM (approved):

mdbooth:
I don't think it's going to be idempotent, but that's much less important that fixing the immediate problem. We can iterate on it.

mdbooth (approved):

EmilienM:
this PR broke ansible-lint, I'm not sure how it wasn't run.

PR #102, opened 2021-06-09
Bridging improvements

EmilienM:
- Switch bridge MTU to 1400 if tunnels are being used.
When using VXLAN tunnels between the bridges, it's important to set proper MTU because the
Ethernet frame is being encapsulated in `Ethernet + IP + UDP + Overlay Header`.
This means that the Maximum Transmission Unit for the underlay needs to be adjusted to something close to 1400.
Using 1400 is a safe number for now which should give up flexibility in
using IPv4 and IPv6 in the future.

- Remove default value for tunnel_remote_ips in template
The default value for tunnel_remote_ips is already sets in default vars,
we don't want to duplicate it in the template.

### DIFF
--- a/playbooks/templates/dev-install_net_config.yaml.j2
+++ b/playbooks/templates/dev-install_net_config.yaml.j2
@@ -26,7 +26,7 @@ network_config:
   - type: interface
     name: dummy0
     nm_controlled: true
-{% for ip in tunnel_remote_ips | default([]) %}
+{% for ip in tunnel_remote_ips %}
   - type: ovs_tunnel
     name: "tun-ctlplane-{{ ip | to_uuid }}"
     tunnel_type: vxlan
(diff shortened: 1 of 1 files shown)


mandre (commented):
/lgtm

EmilienM:
I'm holding it a bit, I just realized a potential issue, I'm digging this out.

EmilienM:
@mdbooth this one is ready for review.

mandre (approved):
Approving again since nothing has changed since I last reviewed. Is the potential issue you mentioned earlier not relevant anymore?

EmilienM:
> Approving again since nothing has changed since I last reviewed. Is the potential issue you mentioned earlier not relevant anymore?

this is ready to be merged I think.
Thanks

mdbooth (approved):
I wonder if we might want to either make the 2 MTUs configurable, or discover the real MTU somehow and subtract the encapsulation overhead from that. However, until we find a reason to do that, this /lgtm

EmilienM:
> I wonder if we might want to either make the 2 MTUs configurable, or discover the real MTU somehow and subtract the encapsulation overhead from that. However, until we find a reason to do that, this /lgtm

ack, I'll keep that in mind for later probably. Thanks.

PR #103, opened 2021-06-09
Remove rhceph-4-osd-for-rhel-8-x86_64-rpms from default OSP repos

EmilienM:
This shouldn't be needed because we pull ceph-ansible from
rhceph-4-tools-for-rhel-8-x86_64-rpms and then everything else is
containerized.

I'm removing it because the repo isn't available in some RHSM
subscriptions for some reason. I'll propose a revert if I find out that
this patch caused more pain.


mandre (approved):
/lgtm

PR #104, opened 2021-06-10
Workaround to make Manila working on OSP16.1

EmilienM:
We need to manually add the CephNfs to the registry because it's
missing in the Standalone role for OSP16.1, there is a missing backport
from upstream to downstream:
https://review.opendev.org/c/openstack/tripleo-heat-templates/+/773205/

While this is being fixed via BZ#1969962, let's add the service
ourselves, which won't hurt any deployment (even master) since the list
will simply merge if the service already exists (which is the case on
> 16.1).

### DIFF
--- a/playbooks/install_stack.yaml
+++ b/playbooks/install_stack.yaml
@@ -151,6 +151,10 @@
     set_fact:
       sriov_services_fact: "{{ sriov_interface|default(None)| ternary(sriov_services, []) }}"
 
+  - name: Set fact for Manila services overrides
+    set_fact:
+      manila_services_fact: "{{ manila_enabled | ternary(manila_services, []) }}"
+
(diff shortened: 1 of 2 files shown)


mandre (approved):
Don't we also need to set `ManilaCephFSCephFSProtocolHelperType: 'NFS'` in the environment like what https://review.opendev.org/c/openstack/tripleo-heat-templates/+/773205 does?

EmilienM:
> Don't we also need to set `ManilaCephFSCephFSProtocolHelperType: 'NFS'` in the environment like what https://review.opendev.org/c/openstack/tripleo-heat-templates/+/773205 does?

not needed since we load /usr/share/openstack-tripleo-heat-templates/environments/manila-cephfsganesha-config.yaml which sets it for us already.

PR #105, opened 2021-06-10
Don't configure CephClusterName on central node

EmilienM:
Due to BZ#1901143 and BZ#1903005, we can't override the Ceph cluster
name where Manila is deployed.

In the case of our Edge setup, we only deploy Manila on the central
site so let's skip this parameter there and configure only the AZ
nodes.

### DIFF
--- a/playbooks/templates/standalone_parameters.yaml.j2
+++ b/playbooks/templates/standalone_parameters.yaml.j2
@@ -141,8 +141,10 @@ parameter_defaults:
   GlanceBackendID: central
   GlanceEnabledImportMethods: web-download,copy-image
   GlanceStoreDescription: 'central rbd glance store'
-{% endif %}
+{% else %}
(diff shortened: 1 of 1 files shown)


mandre (approved):
/lgtm

playbooks/templates/standalone_parameters.yaml.j2
```diff
@@ -141,8 +141,10 @@ parameter_defaults:
   GlanceBackendID: central
   GlanceEnabledImportMethods: web-download,copy-image
   GlanceStoreDescription: 'central rbd glance store'
-{% endif %}
+{% else %}
+  # BZ#1901143 and BZ#1903005
```
mdbooth, line 145: 🙏

Every time a developer leaves a comment containing bug link next to a hacky workaround, a fairy gets its wings.

PR #106, opened 2021-06-15
Switch `yum` to `package` module

EmilienM:
We should use the `package` module instead of `yum` to avoid
compatibility issues on RHEL systems.


EmilienM:
Not needed apparently, will re-open if I hit the issue again.

PR #108, opened 2021-06-16
Rename Manila share type

EmilienM:
To make it easier to discover, let's rename the type that will be used
for Manila CSI testing, to be `default`

Technically it changes nothing but it'll make it so the Manila CSI
operator will create a storage class with the name
`csi-manila-default`; which is more generic than
`csi-manila-cephfsnfstype`.

More context: https://github.com/openshift/csi-driver-manila-operator/pull/103


mandre (approved):
:shipit:

PR #109, opened 2021-06-18
Allow to disable swift for the openshift user

EmilienM:
Add a new option `swiftoperator_enabled` (true by default).

It helps to configure whether or not we want to add the swiftoperator role to
the openshift user in openshift project (in Keystone).
Set this to false and the openshift user won't have access
to Swift. It can be useful if you want to force OpenShift to
deploy the Image registry in Cinder instead of Swift.


EmilienM:
cc @MaysaMacedo

mandre (approved):
Thanks ;-)

PR #110, opened 2021-06-18
Add working example of overrides for tripleomaster deployment

EmilienM:
If we want to deploy TripleO master, we need to use current-tripleo-dev
repo and not current-tripleo, let's just document it here in an example
that was tested and worked fine.


mandre (approved):
/lgtm

PR #112, opened 2021-06-23
Add a condition to reboot the node after deployment

EmilienM:
If we deploy OSP with Kernel args, we need to reboot. It's possible to
configure Kernel args without SR-IOV, so let's add a condition, that
will reboot if we use dpdk_kernel_args param.


mandre (commented):
/lgtm

PR #114, opened 2021-06-24
Allow to deploy OSP with CPU partitioning without SR-IOV

EmilienM:
If we set the `dpdk_kernel_args` parameter but don't want SR-IOV to be
configured, we can now do it with that patch, since we separate the
list of services for DPDK & SR-IOV.

### DIFF
--- a/playbooks/install_stack.yaml
+++ b/playbooks/install_stack.yaml
@@ -177,6 +177,10 @@
     set_fact:
       sriov_services_fact: "{{ sriov_interface|default(None)| ternary(sriov_services, []) }}"
 
+  - name: Set fact for DPDK services overrides
+    set_fact:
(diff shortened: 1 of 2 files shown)


playbooks/install_stack.yaml
```diff
@@ -393,8 +398,7 @@
     script: files/export-dcn.sh
 
   - name: Reboot if SR-IOV is enabled (to apply kernel changes)
-    when:
-      - sriov_interface is defined or dpdk_kernel_args is defined
+    when: false
```
mandre, line 401: Hmm, interesting. Did you really intend to change this line?

EmilienM: yeah I'm debugging something. the PR is in draft for now.

mandre (commented):
/lgtm

PR #115, opened 2021-06-28
Drop unneeded double quotes

EmilienM:
These double quotes can be problematic, let's remove them.

### DIFF
--- a/playbooks/prepare_host.yaml
+++ b/playbooks/prepare_host.yaml
@@ -35,7 +35,7 @@
       - name: Install virt module
         shell: |
           dnf module disable -y virt:rhel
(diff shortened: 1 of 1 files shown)


playbooks/prepare_host.yaml
```diff
@@ -35,7 +35,7 @@
       - name: Install virt module
         shell: |
           dnf module disable -y virt:rhel
-          dnf module enable -y virt:"{{ rhsm_release }}"
+          dnf module enable -y virt:{{ rhsm_release }}
```
mdbooth, line 38: I guess it doesn't matter as long as rhsm_release is a single word, which it always should be.

I wouldn't want to rely on the lack of escaping here to allow passing other arguments, for eg.

PR #116, opened 2021-06-28
prepare_host: only reboot if actually needed

EmilienM:
use the `needs-restarting` tool to figure out if a server actually
needs a reboot. In some cases, it doesn't (ie the kernel is already up
to date, etc), so we'll save deployment time.


mandre (commented):
/lgtm

PR #118, opened 2021-06-29
Remove DPDK variables

EmilienM:
They are too rigid, let's just keep the kernel_args for now.
Users can still use extra_heat_params for their DPDK configs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,8 +102,8 @@ There is no need to mount /var/lib/cinder and /var/lib/nova separately if / is l
 
 ### NFV enablement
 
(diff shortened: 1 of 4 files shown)


mdbooth (approved):
It's a reasonable simplification. I'd be interested to know if we could reasonably automate, or at least massively simplify, the calculation of isolated CPUs.

EmilienM:
> It's a reasonable simplification. I'd be interested to know if we could reasonably automate, or at least massively simplify, the calculation of isolated CPUs.

@christophefontaine has some magics I think :)

PR #119, opened 2021-07-01
prepare_host: fix needs-reboot task

EmilienM:
The task will be marked as failed when the reboot is needed, this patch
fixes that by ignoring the error. It is fine, because a non-zero return
code means a reboot is needed.


mandre (commented):
/lgtm

PR #120, opened 2021-07-02
install_stack: Add low memory option

christophefontaine:
Enables the env file /usr/share/openstack-tripleo-heat-templates/environments/low-memory-usage.yaml,
this reduces the RAM usage from 16GB to 6GB.

### DIFF
--- a/playbooks/vars/defaults.yaml
+++ b/playbooks/vars/defaults.yaml
@@ -76,6 +76,10 @@ standalone_role: /usr/share/openstack-tripleo-heat-templates/roles/Standalone.ya
 standalone_role_overrides: []
 standalone_extra_config: {}
 
+# Sets the number of Neutron/Nova/... workers to 1
+# Uses the environement file low-memory-usage.yaml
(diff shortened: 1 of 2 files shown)


playbooks/vars/defaults.yaml
```diff
@@ -76,6 +76,10 @@ standalone_role: /usr/share/openstack-tripleo-heat-templates/roles/Standalone.ya
 standalone_role_overrides: []
 standalone_extra_config: {}
 
+# Sets the number of Neutron/Nova/... workers to 1
+# Uses the environement file low-memory-usage.yaml
+low_memory_usage: false
```
mdbooth, line 81: Is this a candidate for being enabled by default?

EmilienM: I think it has the downside that the services will be slower which I don't think we want. Keep in mind that the major use case now is deploying OCP on top of OSP, which requires beefy resources, so I think we're fine with it disabled by default for now.

PR #122, opened 2021-07-05
Only keep Heat running when deploying DCN/central server

EmilienM:
We don't need (nor want) to keep the ephemeral Heat services running
once the deployment is finished, except when deploying a central site,
since we'll collect Heat stack data later.
Otherwise, someone who'll try to redeploy will have to kill Heat
manually, which isn't a great user experience.

### DIFF
--- a/playbooks/install_stack.yaml
+++ b/playbooks/install_stack.yaml
@@ -380,6 +380,11 @@
     become: true
     become_user: root
 
+  - name: Create fact for heat_keep_running
+    set_fact:
+      heat_keep_running: true
+    when: dcn_az == "central"
+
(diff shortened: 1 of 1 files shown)


mdbooth:
@gprocunier was talking about a method for re-running tripleo-deploy with new config, which I suspect relies on Heat still running? Would this prevent this from working?

mdbooth:
@EmilienM Here: https://github.com/shiftstack/dev-install/pull/123

EmilienM:
> @gprocunier was talking about a method for re-running tripleo-deploy with new config, which I suspect relies on Heat still running? Would this prevent this from working?

yes, indeed.

PR #124, opened 2021-07-08
Save one reboot when using SR-IOV or Kernel args

EmilienM:
We do not need to reboot after upgrading the rpms if we deploy TripleO
with SR-IOV or/and Kernel args, because we reboot anyway right after
TripleO being deployed. We'll save one reboot.


mandre (commented):
/lgtm

PR #125, opened 2021-07-09
Add support for cephadm instead of ceph-ansible

EmilienM:
Ceph has given up ceph-ansible and therefore TripleO had to remove its
support going forward.

This patch will detect if ceph-ansible can still be used and if not
it'll configure ceph using `cephadm`.

From a user standpoint, dev-install still offers 3 options:

* Disable the management of Ceph from dev-install; if users don't want
  Ceph or if they want to use `extra_heat_params`.
* Let cephadm to install Ceph OSDs on available disks that aren't
  rotational (for performance purposes). This is different from
  ceph-ansible where we could choose the disks, but one can achieve it
  via option 1 if they don't have SSD/NVME disks.
* Let dev-install creating LVM resources on a loop device and let
  cephadm to use these volumes to configure Ceph. Note that cephadm
  doesn't need the DB volume but this shouldn't matter anyway, as the
  data volume should be big enough to run the workloads (volumes, vms,
  etc), since it's 96% of the total size of the loop device.

In other words, this change is backward compatible except if you
deployed Ceph with `ceph_devices` that are rotational; then no OSD will
be deployed and cephadm will crash. In that case it's suggested to go
with option 1 previously described, but dev-install maintainers don't
suggest to install Ceph on these disks because of performance impact.

### DIFF
--- a/playbooks/install_stack.yaml
+++ b/playbooks/install_stack.yaml
@@ -153,10 +153,40 @@
   - name: Install the tripleo client
     yum:
       name:
-      - ceph-ansible
       - python3-tripleoclient
     become: true
     become_user: root
+
+  - name: Prepare the deployment for Ceph
+    when: ceph_enabled
+    block:
+      - name: Check if ceph-ansible can be used to deploy Ceph
+        stat:
+          path: /usr/share/openstack-tripleo-heat-templates/environments/ceph-ansible
+        register: st_ceph
+      - name: Create ceph_ansible_enabled fact
+        set_fact:
+          ceph_ansible_enabled: "{{ st_ceph.stat.isdir is defined and st_ceph.stat.isdir }}"
+      - name: Show a message when ceph_devices is used but empty
+        debug:
+          msg: >
+            ceph_devices contains a list of disks but cephadm is going to use all available SSD/NVME disks.
+            Set the param to true to skip this message or remove it if you don't have available SSD/NVME disks.
+        when:
+          - not ceph_ansible_enabled
+          - ceph_devices is defined
+          - ceph_devices is not true
+          - (ceph_devices | length) > 0
+      - name: Create ceph facts
+        set_fact:
+          ceph_env_base: "{{ ceph_ansible_enabled | ternary('ceph-ansible', 'cephadm') }}"
+          ceph_env_name: "{{ ceph_ansible_enabled | ternary('ceph-ansible', 'cephadm-rbd-only') }}"
+      - name: "Install {{ ceph_package }}"
+        yum:
+          name:
+          - "{{ ceph_env_base }}"
+        become: true
+        become_user: root
 
   - name: Create dev-install_net_config.yaml
     template:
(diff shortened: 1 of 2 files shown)


EmilienM:
Highly WIP, don't review it yet please.

EmilienM:
I tested it against real disks and also with the default (loop device) and it worked fine. I think this is good to go unless there are concerns about the gotchas in the commit message.

playbooks/install_stack.yaml
```diff
@@ -153,11 +153,38 @@
...
+  - name: Prepare the deployment for Ceph
+    when: ceph_enabled
+    block:
+      - name: Check if ceph-ansible can be used to deploy Ceph
+        stat:
+          path: /usr/share/openstack-tripleo-heat-templates/environments/ceph-ansible
+        register: st_ceph
+      - name: Create ceph_ansible_enabled fact
+        set_fact:
+          ceph_ansible_enabled: "{{ st_ceph.stat.isdir is defined and st_ceph.stat.isdir | ternary(true, false) }}"
```
mdbooth, line 169: nit: I'm pretty sure `| ternary(true, false)` is a noop and can just be omitted.

playbooks/install_stack.yaml
```diff
@@ -153,11 +153,38 @@
...
+        become: true
+        become_user: root
+        when: ceph_ansible_enabled
+      - name: Install cephadm
+        yum:
+          name:
+          - cephadm
+        become: true
+        become_user: root
+        when: not ceph_ansible_enabled
```
mdbooth, line 186: I don't think you need the conditionals here because you've already got `ceph_env_name`:
```yaml
- name: Install {{ ceph_env_name }}
  yum:
    name:
      {{ ceph_env_name }}
  become: true
  become_user: root
```

playbooks/install_stack.yaml
```diff
@@ -331,7 +358,7 @@
       service_envs: "{{ service_envs | union(ceph_env) }}"
     vars:
       ceph_env:
-        - /usr/share/openstack-tripleo-heat-templates/environments/ceph-ansible/ceph-ansible.yaml
+        - "/usr/share/openstack-tripleo-heat-templates/environments/{{ ceph_env_name }}/{{ ceph_env_name }}.yaml"
```
mdbooth, line 361: I assume `cephadm/cephadm.yaml` exists?

mdbooth (commented):
Nice work! Couple of cleanups in there if you feel like it, not important.

I've checked that I don't think it breaks ceph-ansible. Obviously don't know about cephadm. Haven't tested.

playbooks/install_stack.yaml
EmilienM: https://github.com/openstack/tripleo-heat-templates/blob/master/environments/cephadm/cephadm.yaml

mandre:
Looks good. Perhaps we should print a warning when `ceph_devices` is set and we detect that we'll use cephadm to deploy ceph, to let the user know this variable WILL NOT be honored.

EmilienM:
> Looks good. Perhaps we should print a warning when `ceph_devices` is set and we detect that we'll use cephadm to deploy ceph, to let the user know this variable WILL NOT be honored.

done

mandre:
> > Looks good. Perhaps we should print a warning when `ceph_devices` is set and we detect that we'll use cephadm to deploy ceph, to let the user know this variable WILL NOT be honored.
> 
> done

Perhaps I've completely misunderstood the gotcha with `ceph_devices`, but it seems to me that it's not used unless the environment uses ceph-ansible. In this case the warning you've added doesn't really address my concern. I want to avoid:
- users setting `ceph_devices` like before and expecting it to work with cephadm
- the cephadm crash reported in the PR description (how is it possible if `ceph_devices` is ignored with cephadm?)

mandre (commented):
LGTM

PR #129, opened 2021-07-15
External network to be shared

zerodayz:
If you are deploying by openshift project, the external network needs to be shared. So you can later create a floating ips for the openshift installer.


mdbooth:
/hold

Re-testing this one in case it was down to local config changes. This should not be required.

zerodayz:
It actually works without sharing the network ;-)

PR #133, opened 2021-07-30
prepare_host: add task to make sure persistent NIC is enabled

EmilienM:
It's possible that CentOS or RHEL images come with persistent NIC
naming disabled and we don't want that because it'll cause random
networking issues after reboots.

Let's make sure we remove that argument from the grub cmd and rebuild
grub.

### DIFF
--- a/playbooks/prepare_host.yaml
+++ b/playbooks/prepare_host.yaml
@@ -20,6 +20,16 @@
       key: "{{ item }}"
     loop: "{{ authorized_keys }}"
 
+  - name: Make sure persistent NIC names is not disabled # noqa 301
+    shell: |
+      if grep -q net.ifnames=0 /etc/default/grub; then
+          sed -i 's/net.ifnames=0 //g' /etc/default/grub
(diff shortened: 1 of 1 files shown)


mdbooth (commented):
LGTM. Before I approve it, did you test it manually?

EmilienM:
yes except that I'm not entirely sure that we don't need to reboot right after. but honestly we already reboot a few tasks later so I guess this is fine now.

mdbooth:
Now that I think about it, this is presumably going to rename nics. So anywhere we refer to nic names is going to potentially break after this? Perhaps we do need that immediate reboot. Is that what you were thinking of?

EmilienM:
> Now that I think about it, this is presumably going to rename nics. So anywhere we refer to nic names is going to potentially break after this? Perhaps we do need that immediate reboot. Is that what you were thinking of?

yes exactly. I'll add a reboot, just to be safe.

mdbooth (approved):
/lgtm

PR #134, opened 2021-07-30
Safe defaults to install OSP 16.1 and 16.2

EmilienM:
## Adding example config for deploying 16.2 beta
* Add an example of 16.2-beta config.
* Introduce `virt_release` to override the default virt:rhel module that is installed. Note that the virt module needs to be set to `av` on RHEL 8.4 and onward.

## Improve defaults for 16.1 when using puddles
16.2 is not GA'ed yet so it's safer to use 16.1 when downloading content from puddles. Content from 16.2 might be unstable.

### DIFF
--- a/example-overrides/local-overrides-osp16-2.yaml
+++ b/example-overrides/local-overrides-osp16-2.yaml
@@ -1,4 +1,42 @@
 standalone_host: <standalone FQDN>
 public_api: <IP address used to reach the node>
-# Get latest 16.2 puddle
-rhos_release: 16.2
+rhsm_enabled: true
+redhat_registry_credentials: "'<user>': '<password>'"
+rhsm_org_id: "<ID>"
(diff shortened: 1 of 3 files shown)


mandre (commented):
IMO, we should improve the commit message as it doesn't completely reflects the change.
We're changing the default to 16.1, and providing an example overrides file to work with 16.2 beta on RHEL 8.4.

EmilienM:
@mandre done. Thanks for the feedback!

mandre:
Thanks, I see you've updated the PR description but it's still showing the same commit. Did you perhaps forgot to force push?

EmilienM:
> Thanks, I see you've updated the PR description but it's still showing the same commit. Did you perhaps forgot to force push?

yes I forgot to push -f :-)

EmilienM:
Thanks!

PR #135, opened 2021-08-04
PCI passthrough local_override.yaml example

adduarte:
All the necessary parameters to configured pci passthrough can be specified in the local_override.yaml file

### DIFF
--- a/example-overrides/local-overrides-osp16-2-with-pcipassthrough.yaml
+++ b/example-overrides/local-overrides-osp16-2-with-pcipassthrough.yaml
@@ -1,0 +1,23 @@
+standalone_host: <standalone FQDN>
+public_api: <IP address used to reach the node>
+# Get latest 16.2 puddle
+rhos_release: 16.2
+
+
+# PCIPassthrough example for device " 0b:00.0 VGA compatible controller [0300]: Matrox Electronics Systems Ltd. G200eR2 [102b:0534] (rev 01)"
+# For available device_type values see https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/16.1/html/configuring_the_compute_service_for_instance_creation/configuring-pci-passthrough
+standalone_extra_config:
+  nova::pci::aliases:
+  - name: "VGAalias"
+    vendor_id: "102b"
+    product_id: "0534"
+    device_type: "type-PCI"
+
+extra_heat_params:
+  NovaSchedulerDefaultFilters: ['RetryFilter','AvailabilityZoneFilter','ComputeFilter','ComputeCapabilitiesFilter','ImagePropertiesFilter','ServerGroupAntiAffinityFilter','ServerGroupAffinityFilter','PciPassthroughFilter','NUMATopologyFilter']
+  NovaSchedulerAvailableFilters: ["nova.scheduler.filters.all_filters","nova.scheduler.filters.pci_passthrough_filter.PciPassthroughFilter"]
+  NovaPCIPassthrough:
+    - vendor_id: "102b"
+      product_id: "0534"
+
+kernel_args: "iommu=pt intel_iommu=on"


mdbooth:
/lgtm

We might refine this if the need arises, but I'm happy to wait to see if that happens.

playbooks/vars/defaults.yaml
```diff
@@ -145,6 +145,16 @@ kernel_services:
 #sriov_nic_numvfs:
 #sriov_nova_pci_passthrough:
 
+
+# For pci passthrough provide the relevant fields. for example:
+#pci_passthrough_devices:
```
EmilienM, line 150: this is not enough, when configuring NovaPCIPassthrough you need to add a filter to Nova, which is done via /usr/share/openstack-tripleo-heat-templates/environments/services/neutron-ovn-sriov.yaml (automatically loaded when setting `sriov_interface`).

So maybe we could add a conditional in `Add sriov to enabled services` in `playbooks/install_stack.yaml` to automatically load that file, so we get the right Filters configured in Nova. However, it'll also configure Neutron ML2 mechanism driver to ['sriovnicswitch','ovn'], I don't know if this is a problem if we don't configure SR-IOV interface.

playbooks/templates/standalone_parameters.yaml.j2
```diff
@@ -59,7 +59,10 @@ parameter_defaults:
 {% if sriov_interface is defined %}
   NeutronPhysicalDevMappings: "hostonly:{{ sriov_interface }}"
   NovaPCIPassthrough: {{sriov_nova_pci_passthrough | mandatory | to_json }}
+{% elif pci_passthrough_devices is defined %}
```
EmilienM, line 62: Why don't we re-use sriov_interface? What is configured with `sriov_interface`, that we don't want with `pci_passthrough_devices` ?

EmilienM (changes requested):

playbooks/templates/standalone_parameters.yaml.j2
EmilienM: we clarified it, and this is good to me.

adduarte:
/hold
Turns out we can proably do all of this by just including the correct fields into standalone_parameters.yaml
like so: 
 #PCIPASSTHROUGH TEST
  NovaSchedulerDefaultFilters: ['RetryFilter','AvailabilityZoneFilter','ComputeFilter','ComputeCapabilitiesFilter','ImagePropertiesFilter','ServerGroupAntiAffinityFilter','ServerGroupAffinityFilter','PciPassthroughFilter','NUMATopologyFilter']
  NovaSchedulerAvailableFilters: ["nova.scheduler.filters.all_filters","nova.scheduler.filters.pci_passthrough_filter.PciPassthroughFilter"]
  nova::pci::aliases: [{"name": "VGAalias","vendor_id": "102b","product_id": "0534","devic_type": "type-PCI"}]

  #0b:00.0 VGA compatible controller [0300]: Matrox Electronics Systems Ltd. G200eR2 [102b:0534] (rev 01)
  NovaPCIPassthrough: [{"vendor_id": "102b", "product_id": "0534"}]
  KernelArgs: "iommu=pt intel_iommu=on"

playbooks/vars/defaults.yaml
adduarte: right this is not enough, thus why the WIP in the title.

adduarte: This no longer applies

example-overrides/local-overrides-osp16-2-with-pcipassthrough.yaml
```diff
@@ -0,0 +1,25 @@
...
+# PCIPassthrough example for device " 0b:00.0 VGA compatible controller [0300]: Matrox Electronics Systems Ltd. G200eR2 [102b:0534] (rev 01)"
+# For available device_type values see https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/16.1/html/configuring_the_compute_service_for_instance_creation/configuring-pci-passthrough
+standalone_extra_config:
+  nova::pci::aliases:
+  - name: "VGAalias"
+    vendor_id: "102b"
+    product_id: "0534"
+    device_type: "type-PCI"
+
+NovaPCIPassthrough:
```
EmilienM, line 16: you need to move that under `extra_heat_params`.

PR #137, opened 2021-08-04
Combined improvements for Manila

EmilienM:
## Add required variables for Ganesha to listen on public network

A few months ago we decided that Manila shares would listen on public
network, because we don't have a storage network in Standalone.

For that to happen, we need to configure Ganesha to listen on the
public network (and not the control plane), so we need to configuring
the following:

* ganesha_vip: for pacemaker to create the VIP
* ceph_nfs_bind_addr: for ceph-ansible, taking care of configuring
  ganesha with the public IP
* tripleo_cephadm_ceph_nfs_bind_addr: for cephadm, taking care via
  tripleo-ansible of configuring ganesha with the public IP.

## Increase default quotas for Manila
50 shares doesn't seem enough when running the CSI test suite in parallel mode.
Let's increase it to 150 shares and 150 snapshots

### DIFF
--- a/playbooks/install_stack.yaml
+++ b/playbooks/install_stack.yaml
@@ -263,7 +263,12 @@
         standalone_extra_config: "{{ standalone_extra_config | combine(manila_extra_config) }}"
       vars:
         manila_extra_config:
+          # Needed by Pacemaker to create the VIP
           ganesha_vip: "{{ public_api }}"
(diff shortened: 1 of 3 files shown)


EmilienM:
/hold I need to test it

FYI @tombarron

EmilienM:
@mandre tested, ready

mandre (commented):
LGTM

PR #141, opened 2021-08-18
Configure Cinder API to allow users to set volume states

EmilienM:
By default, volume states can only be changed by admin but in
dev-install cases, we are operating our cloud and sometimes need to
reset the state of a volume.

e.g. in OpenShift CI, sometimes volumes aren't cleaned correctly and we
have to manually clear them up. Thanks to this patch, the tenant will
be able to set the state of their own volumes.


EmilienM:
It's ready for review and I applied this change on `mecha`.

pierreprinetti:
That’s useful. Thanks!

/approve
/lgtm

pierreprinetti (approved):

PR #142, opened 2021-08-27
Ignore error if yum/dnf fails on localhost

dulek:
We shouldn't assume the script is run on something having yum/dnf. Let's
ignore the error of the task requiring it and assume that user can
install it manually if it's not there.

Also it prevents from errors like below when user running the playbook
isn't added to `wheel` group. You can fix it with `sudo gpasswd -a
<user> wheel` BTW.

fatal: [standalone]: FAILED! => {"changed": false, "msg": ["Could not
detect which major revision of yum is in use, which is required to
determine module backend.", "You can manually specify
use_backend to tell the module whether to use the yum (yum3) or dnf
(yum4) backend})"]}

### DIFF
--- a/playbooks/roles/simpleca/tasks/main.yaml
+++ b/playbooks/roles/simpleca/tasks/main.yaml
@@ -4,6 +4,8 @@
 - name: install crypto library for openssl_* modules
   become: true
   become_user: root
+  # To not fail on non-RPM systems let's just ignore errors.
+  ignore_errors: true
   yum:
     state: installed
     name: python3-cryptography


EmilienM:
@dulek oh I remember now, you're probably using Silverblue, which is a containerized distro, so you can't use yum/dnf/rpm.
Please do not remove this task and make it ignoring errors, with `ignore_errors: true`. Your system probably have the package installed, but think about systems which don't have it, they'll need it to generate SSL bits.

Thanks

dulek:
> @dulek oh I remember now, you're probably using Silverblue, which is a containerized distro, so you can't use yum/dnf/rpm.
> Please do not remove this task and make it ignoring errors, with `ignore_errors: true`. Your system probably have the package installed, but think about systems which don't have it, they'll need it to generate SSL bits.
> 
> Thanks

Ouch, this thing runs on my laptop? I thought it's on the host, I should be able to figure it out locally then! And it explains why it worked for me once. 

I use regular Fedora 34, not Silverblue.

dulek:
Apparently I need to run this in order to fix it: `sudo gpasswd -a <laptop-user> wheel`.

dulek:
Anyway there's value in adding `ignore_errors: true` - @gryf uses Gentoo so no yum/dnf there anyway.

PR #144, opened 2021-09-01
Only deploy useful tuned profiles

EmilienM:
We don't need to deploy them all, it can actually causes some packaging
dependencies problems on OSP 16.1.


mandre (commented):
/lgtm

PR #147, opened 2021-09-04
 [WIP] Enable ovsdpdk

adduarte:


playbooks/install_stack.yaml
```diff
@@ -370,9 +374,7 @@
   - name: Install tuned profiles useful for OSP
     yum:
       name:
-      - tuned
-      - tuned-profiles-cpu-partitioning
-      - tuned-profiles-realtime
+      - tuned-profiles-*
```
adduarte, line 377: had to change this back because otherwise I get error about tuned-profiles-realtime not being available. 
This should be changed back before merging.

EmilienM: yeah I made that change because the packaging on RHEL 8.4 / OSP 16.2 was broken for tuned-*, and it was trying to install mixed versions of tuned. Maybe it's solved now, that RC is out. To be tested again.

playbooks/vars/defaults.yaml
```diff
@@ -138,13 +138,20 @@ update_local_pki: false
 sriov_services:
   - OS::TripleO::Services::NeutronSriovAgent
 
+ovsdpdk_services:
+  - OS::TripleO::Services::ComputeNeutronOvsDpdk
+  - OS::TripleO::Services::OvsDpdkNetcontrold
```
EmilienM, line 143: This shouldn't be used, netcontrold is not used anymore and its features were added to native OVS.

playbooks/install_stack.yaml
EmilienM: we should use cpu partitioning anyway, I think you could remove that change.

playbooks/vars/defaults.yaml
```diff
@@ -138,13 +138,20 @@ update_local_pki: false
...
+
 kernel_services:
   - OS::TripleO::Services::BootParams
 
 #sriov_interface:
 #sriov_nic_numvfs:
 #sriov_nova_pci_passthrough:
 
+#ovsdpdk
+ovsdpdk: true
```
EmilienM, line 153:
```suggestion
dpdk_enabled: false
```

playbooks/vars/defaults.yaml
EmilienM:
```suggestion
```

playbooks/install_stack.yaml
```diff
@@ -384,6 +386,14 @@
         - /usr/share/openstack-tripleo-heat-templates/environments/services/neutron-ovn-sriov.yaml
     when: sriov_interface is defined
 
+  - name: Add ovsdpdk to enabled services
+    set_fact:
+      service_envs: "{{ service_envs | union(ovsdpdk_env) }}"
+    vars:
+      ovsdpdk_env:
+        - /usr/share/openstack-tripleo-heat-templates/environments/services/neutron-ovs-dpdk.yaml
```
EmilienM, line 394: we should have two blocks here, one for when using OVS and another one when using OVN.

on https://github.com/shiftstack/dev-install/pull/148 I'm adding `ovn_enabled` param, we could probably use it here.

EmilienM:
@adduarte FYI I reworked the PR in https://github.com/shiftstack/dev-install/pull/152

PR #148, opened 2021-09-08
Add support for OVS HW offload

EmilienM:
* Allow to deploy SR-IOV without OVN (but with OVS), with
  `ovn_enabled`, set to True by default.
* Add an option `ovs_hw_offload` which when set to True will configure
  OVS Hardware offload parameters and also switch the SR-IOV nic to
  switchdev.
* Add an example of overrides that can be used to deploy OVS HW
  offload.


example-overrides/local-overrides-osp16-2-with-pcipassthrough-ovs-hw-offload.yaml
```diff
@@ -0,0 +1,19 @@
...
+# with SR-IOV passthrough for enp130s0f0 NIC, with hugepages and CPU partitioning.
+standalone_host: <standalone FQDN>
+public_api: <IP address used to reach the node>
+rhos_release: 16.2
+sriov_interface: enp130s0f0
+sriov_nic_numvfs: 63
+sriov_nova_pci_passthrough:
+  - devname: "enp130s0f0"
+    physical_network: "hostonly"
+kernel_args: "default_hugepagesz=1GB hugepagesz=1G hugepages=128 iommu=pt intel_iommu=on isolcpus=4-47"
```
mdbooth, line 11: Do you want VMs to use these hugepages? If so you'll need to update the OpenStack flavors, too.

EmilienM: yeah the flavors are created/updated manually for now, I guess this could be a new (separate) feature to dev-install. Although too opinionated in my taste. Because there are many ways of configuring the flavor.

playbooks/templates/standalone_parameters.yaml.j2
```diff
@@ -58,7 +58,9 @@ parameter_defaults:
   NeutronFlatNetworks: {{ neutron_flat_networks }}
 {% if sriov_interface is defined %}
   OVNCMSOptions: "enable-chassis-as-gw"
+{% if neutron_physical_dev_mappings is not defined %}
   NeutronPhysicalDevMappings: "hostonly-sriov:{{ sriov_interface }}"
+{% endif %}
```
mandre, line 63: Wasn't the intention to have `neutron_physical_dev_mappings` overwrite the setting?

In which case, perhaps something like:
```suggestion
{% if neutron_physical_dev_mappings is defined %}
  NeutronPhysicalDevMappings: {{ neutron_physical_dev_mappings }}
{% else %}
  NeutronPhysicalDevMappings: "hostonly-sriov:{{ sriov_interface }}"
{% endif %}
```

EmilienM: you're very right, thanks for this review.

mandre (commented):
/lgtm
/approve

PR #149, opened 2021-09-13
Few fixes for OSP 16.2

EmilienM:
## Fix Manila extra config
Fix some extra config for Manila ganesha IP

## Fix for tuned
Tuned RT isn't available from OSP 16.2 and not really needed in
standalone use case.

Installing CPU partitioning should be enough.


mandre (commented):
/lgtm

PR #150, opened 2021-09-14
Fix StandaloneExtraConfig

EmilienM:
In a previous commit, we forgot to add a new condition to configure
StandaloneExtraConfig when Manila is enabled.

This patch will fix it.

### DIFF
--- a/playbooks/templates/standalone_parameters.yaml.j2
+++ b/playbooks/templates/standalone_parameters.yaml.j2
@@ -147,7 +147,7 @@ parameter_defaults:
   # over VXLAN.
   NeutronGlobalPhysnetMtu: 1300 
 {% endif %}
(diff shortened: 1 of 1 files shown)


playbooks/templates/standalone_parameters.yaml.j2
```diff
@@ -159,10 +159,12 @@ parameter_defaults:
     # for cephadm support on OSP 17
     tripleo_cephadm_ceph_nfs_bind_addr: "{{ public_api }}"
 {% endif %}
+{% if standalone_extra_config|length > 0
```
mandre, line 162: Did you forget the closing sign?
```suggestion
{% if standalone_extra_config|length > 0 %}
```

mandre (commented):
/lgtm

PR #152, opened 2021-09-23
Add support for SR-IOV + DPDK

EmilienM:
## Create a dedicated SR-IOV network when needed

When SR-IOV is enabled, we will create a dedicated provider network
where the SR-IOV nic will be plugged into.
This will allow to separate the provider networks from regular
interfaces, SR-IOV and later DPDK.

## Add DPDK support

Adding a new parameter, `dpdk_interface` which allows to enable DPDK in
dev-install.

It'll do the following:
* Configure TripleO roles & services, to enable basic parameters for
  OVN or OVS DPDK.
* Manage the provider network dedicated to DPDK.

It won't do the following:
* Configure Tuned profile name
* Configure CPU partitioning for Nova, the host and DPDK
However it provides an example of overrides that can be used to deploy
with DPDK.


EmilienM:
under testing

EmilienM:
I managed to finish a deployment but I still have errors with DPDK: https://paste.opendev.org/show/809591/
I think I'm close. Reviews are welcome.

playbooks/install_stack.yaml
```diff
@@ -387,6 +391,27 @@
...
+    when:
+    - dpdk_interface is defined
+    - ovn_enabled
+
+  - name: Add dpdk-ovs to enabled services
+    set_fact:
+      service_envs: "{{ service_envs | union(dpdk_env) }}"
+    vars:
+      dpdk_env:
+        - /usr/share/openstack-tripleo-heat-templates/environments/services/neutron-ovs.yaml
```
mandre, line 409: Is the `neutron-ovs.yaml` environment file something that we should conditionally import based on the RHOSP version?

Why don't we have a similar environment file for neutron-ovn?

playbooks/install_stack.yaml
```diff
@@ -456,7 +481,7 @@
   - name: Reboot if SR-IOV is enabled (to apply kernel changes)
```
mandre, line 481: It can reboot the node even if SR-IOV is not enabled. I suggest you rename the task:
```suggestion
  - name: Reboot to apply kernel changes
```

playbooks/templates/standalone_parameters.yaml.j2
```diff
@@ -57,9 +57,13 @@ parameter_defaults:
   NeutronBridgeMappings: {{ neutron_bridge_mappings }}
   NeutronFlatNetworks: {{ neutron_flat_networks }}
 {% if sriov_interface is defined %}
-  NeutronPhysicalDevMappings: "hostonly:{{ sriov_interface }}"
+  NeutronPhysicalDevMappings: "hostonly-sriov:{{ sriov_interface }}"
```
mandre, line 60: Don't we also need to map `hostonly-dpdk` to `dpdk_interface` when `dpdk_interface` is defined?

mandre (commented):
A few questions inline. This more or less looks OK to me with my limited knowledge, I'd feel better if someone with better understanding of DPDK reviews this patch as well.

example-overrides/local-overrides-osp16-2-with-dpdk.yaml
```diff
@@ -0,0 +1,56 @@
+standalone_host: <standalone FQDN>
```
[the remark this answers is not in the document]
EmilienM: :+1:

example-overrides/local-overrides-osp16-2-with-dpdk.yaml
```diff
@@ -0,0 +1,56 @@
...
+      ceph_tag: latest
+      name_prefix: openstack-
+      name_suffix: ''
+      namespace: registry.redhat.io/rhosp-rhel8
+      neutron_driver: ovn
+      rhel_containers: false
+      tag: '16.2'
+    tag_from_label: '{version}-{release}'
+
+dpdk_interface: enp130s0f0
```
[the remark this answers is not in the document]
EmilienM: i'll add doc in README

playbooks/install_stack.yaml
EmilienM: because OVN is enabled by default in TripleO starting from OSP 16; so we don't have anything to import by default.

playbooks/templates/standalone_parameters.yaml.j2
EmilienM: @christophefontaine can you please answer?

EmilienM:
I talked to @christophefontaine and I need to change the os-net-config, to do the following when DPDK is enabled:
* Remove br-ex and keep external interface on the system outside of any bridge, to avoid IN/OUT from kernel & userspace, for public traffic (floating IP, etc)
* Do the same for other networks: hostonly, SR-IOV
* Only create a bridge for DPDK
* Update Neutron Bridge mapping in Heat params

EmilienM:
When enabling DPDK, I'm having DHCP issues on non DPDK networks (SR-IOV, hostonly and tenant networks), I'll investigate that before we merge this one.

playbooks/prepare_stack.yaml
```diff
@@ -103,7 +111,9 @@
       fi
     environment:
       OS_CLOUD: standalone
-    when: external_fip_pool_start is defined and external_fip_pool_end is defined
+    when:
+    - external_fip_pool_start is defined and external_fip_pool_end is defined
+    - dpdk_interface is not defined
```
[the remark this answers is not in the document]
EmilienM: done

PR #153, opened 2021-09-23
Add a task to unregister RHSM

EmilienM:
Add a task to unregister RHSM from the system unless `rhsm_ephemeral`
is set to False.

Note that this is set to True by default because most of subscriptions
have limited numbers of systems.

In production or CI environments, you want to set it to False, so the
system can get upgrades on day 2.


mandre (commented):
/lgtm

PR #154, opened 2021-09-27
Switch to OSP 16.2 by default

EmilienM:
16.2 was released, so let's bump to 16.2 by default instead of 16.1.


playbooks/vars/defaults.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-rhos_release: 16.1
```
mandre, line 2: Should we perhaps make a copy of the old `defaults.yaml` to an `example-overrides/local-overrides-osp16-1.yaml` file, if someone ever need to deploy RHOSP16.1?

EmilienM: done

example-overrides/local-overrides-rhel8_2-tripleo-osp16-1.yaml
```diff
@@ -0,0 +1,15 @@
+standalone_host: <standalone FQDN>
```
mandre, line 1: You're missing `rhos_release: 16.1` somewhere.

EmilienM: this is only needed if you want to install from the puddle. Users who want to use the puddle (and not RHSM) would not need all these bits, just to set `rhos_release: 16.1` without RHSM bits.

mandre (commented):
/lgtm

PR #156, opened 2021-10-21
Allow to set admin_password

EmilienM:
In environments we want predictable credentials, let's allow to
override the Keystone admin password which is by default generated
& secured in TripleO.


mandre (commented):
/lgtm

PR #163, opened 2021-10-28
Improve SRIOV integration

EmilienM:
With this patch, I'm able to create a port of type `direct` then boot
a VM from it, using --config-drive=True.

The VM will have its metadata/DHCP provided by Config Drive (not OVN
DHCP) and on the same node we can use the OVS for DPDK integration.


mandre (commented):
/lgtm

PR #169, opened 2021-11-12
Create 'allow_ping' security group

stephenfin:
This complements the 'allow_ssh' security group rule we already create.

### DIFF
--- a/playbooks/prepare_stack.yaml
+++ b/playbooks/prepare_stack.yaml
@@ -146,11 +146,20 @@
     environment:
       OS_CLOUD: standalone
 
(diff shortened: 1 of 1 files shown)


playbooks/prepare_stack.yaml
```diff
@@ -155,6 +155,15 @@
     environment:
       OS_CLOUD: standalone
 
+  - name: Create basic security group which allows ping # noqa 301
+    shell: |
+      if ! openstack security group show allow_ping; then
```
mandre, line 160: We're technically allowing more than just ping but I don't think anyone cares.

mandre (commented):
/lgtm

PR #174, opened 2021-12-07
Wipe disk(s) used for ceph

stephenfin:
If you don't do this, Ceph will moan about the presence of a bootloader. Use of `wipefs` is an alternative to using ceph-volume's `zap` command, which requires ceph be already installed on the host.

### DIFF
--- a/playbooks/prepare_host.yaml
+++ b/playbooks/prepare_host.yaml
@@ -219,7 +219,7 @@
         fstype: xfs
         state: mounted
 
-  - name: Prepare host for Ceph
+  - name: Prepare host for Ceph (loop device)
     when:
     - ceph_enabled
     - ceph_devices is not defined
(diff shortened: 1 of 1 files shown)


stephenfin:
Just tested this. Works as expected :+1:

playbooks/prepare_host.yaml
```diff
@@ -270,6 +270,16 @@
...
         size: 4%VG
+
+  - name: Prepare host for Ceph (disk device)
+    when:
+    - ceph_enabled
+    - ceph_devices is defined
+    block:
+    - name: Wipe filesystem from disk
+      command: wipefs -a "{{ item }}"
+      loop: "{{ ceph_devices }}"
```
EmilienM, line 281: In OSP17, Ceph deployment is done via ceph-adm tool (native to ceph tooling), and we can't provide ceph_devices to it, the tool selects which disk to use for Ceph based on its properties.
Which means that in dev-install we decided to keep supporting `ceph_devices` for backward compatibility (OSP 16, etc), but in OSP17 we recommend to just set it to true; which means that this task will likely fail.

You could do:
``` yaml
- name: Prepare host for Ceph (disk device)
    when:
    - ceph_enabled
    - ceph_devices is defined
    - ceph_devices is not true
    - (ceph_devices | length) > 0
    block:
    - name: Wipe filesystem from disk
      command: wipefs -a "{{ item }}"
      loop: "{{ ceph_devices }}"
```

stephenfin: Done

PR #176, opened 2022-01-10
Fix osc install when running in a virtualenv

gouthampacha:
"--user" shouldn't be specified when installing
pip packages install a python virtualenv. Setting it causes
the following error:

TASK [Install openstack client locally] ****************************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Unsupported parameters for (pip) module: when. Supported parameters include: virtualenv_site_packages, editable, executable, virtualenv, name, extra_args, virtualenv_command, virtualenv_python, state, umask, chdir, requirements, version."}

### DIFF
--- a/playbooks/local_os_client.yaml
+++ b/playbooks/local_os_client.yaml
@@ -120,7 +120,7 @@
   - name: Install openstack client locally
     pip:
       name: python-openstackclient
-      extra_args: --user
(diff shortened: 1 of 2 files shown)


gouthampacha:
ptal @EmilienM @mdbooth

EmilienM:
thanks @gouthampacha !

PR #177, opened 2022-01-20
CentOS 9 support

EmilienM:
* Update README
* Add example overrides for centos9 deployed on tripleo master
* Install modules only for 8 stream. No modules are required on 9


mandre (commented):
/lgtm
/approve

PR #178, opened 2022-01-24
install_stack: disable the use of --standalone arg in recent TripleO

EmilienM:
Tripleoclient removed --standalone option some time ago:
https://review.opendev.org/c/openstack/python-tripleoclient/+/753447

And the TripleO Operator Ansible was updated to not use it anymore:
https://review.opendev.org/c/openstack/tripleo-operator-ansible/+/753853

In dev-install, we now figure out if the version of TripleO that is
being installed on CentOS needs or doesn't need that flag.

Note that we'll have to figure out how to make it work for OSP 17.1 and
OSP 18, since they'll probably contain these changes, but we have time
for now.

### DIFF
--- a/playbooks/install_stack.yaml
+++ b/playbooks/install_stack.yaml
@@ -465,13 +465,23 @@
     become: true
     become_user: root
 
+  - name: Block to figure out if --standalone needs to be passed when deploying tripleo
+    when:
+    - ansible_facts.distribution == 'CentOS'
+    block:
+      - name: Check if the version of TripleO is recent enough to remove --standalone argument
+        set_fact:
+          standalone_toggle: false
+        when:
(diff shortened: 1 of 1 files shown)


playbooks/install_stack.yaml
```diff
@@ -465,13 +465,24 @@
...
+  - name: Block to figure out if --standalone needs to be passed when deploying tripleo
+    when:
+    - ansible_facts.distribution == 'CentOS'
+    block:
+      - name: Check if the version of TripleO is recent enough to remove --standalone argument
+        set_fact:
+          standalone_toggle: false
+        when:
+        - tripleo_repos_branch is defined
+        - tripleo_repos_branch not in ['train', 'ussuri', 'victoria']
```
EmilienM, line 477: oh I missed the case where tripleo_repos_branch is not defined (default, when using master).

EmilienM:
@mandre @mdbooth for review please

mandre (commented):
LGTM

PR #190, opened 2022-05-13
IPv6 Dual Stack on hostonly network

EmilienM:
From now, the `hostonly` network will be configured in dual stack so we
can easily play with IPv6 on that provider network.

We'll use 2001:db8::/64 cidr and make sure that we can create VMs on
that network, and ping from each other and even reach outside out of the
box.

Tested on OSP 16.2.

### DIFF
--- a/playbooks/templates/standalone_parameters.yaml.j2
+++ b/playbooks/templates/standalone_parameters.yaml.j2
@@ -49,6 +49,12 @@ parameter_defaults:
 {% endfor %}
   # needed for vip & pacemaker
   KernelIpNonLocalBind: 1
+  # For OSP16
+  ExtraSysctlSettings:
+    net.ipv6.conf.all.forwarding:
+      value: 1
+  # For OSP17 and beyond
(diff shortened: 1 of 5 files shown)


playbooks/templates/standalone_parameters.yaml.j2
```diff
@@ -49,6 +49,11 @@ parameter_defaults:
 {% endfor %}
   # needed for vip & pacemaker
   KernelIpNonLocalBind: 1
+  # For OSP16
+  ExtraSysctlSettings:
+    net.ipv6.conf.all.forwarding: 1
```
EmilienM, line 54: sounds like this doesn't work

EmilienM: fixed

mdbooth:
The IPv6 subnet is the 'Documentation' range defined in https://www.iana.org/go/rfc3849 👍

mdbooth (approved):

EmilienM:
tested and ready to be merged I think.

PR #191, opened 2022-05-16
ibmcloud: Add variable for primary_nic state and nm_dns management

nhicher:
With this change, user can specify on local-overrides.yaml:

- primary_nic_state to change the required state for the
  interface (eg: primary_nic_state: up)
- set the value for NetworkManager dns config (eg: dns: none)


playbooks/prepare_host.yaml
```diff
@@ -121,15 +120,14 @@
   # resolv.conf
   - name: Enable NetworkManager ownership of resolv.conf
     when:
-      - false
       - ansible_facts.distribution == 'CentOS'
     block:
       - name: Set main/dns=default in NetworkManager
```
mandre, line 125: Perhaps we should update the task name too?
```suggestion
      - name: Set main/dns in NetworkManager
```

mandre (commented):
Hi @nhicher , it would probably be easier if you create a new PR against master. It took me some time to see you created a PR against an open PR. I didn't even know you could do this :D

playbooks/prepare_host.yaml
nhicher: Hello @mandre,
I test dev-install with the branch @EmilienM created for ibm baremetal instance. I have probably some others stuffs to add.
Thanks for the review

EmilienM:
I'm not sure why we needed this just on IBM cloud, I'll close this now. Re-open if needed.

PR #192, opened 2022-06-17
RHEL9 - OSP17 required changes

EmilienM:
To support RHEL9/OSP17, we need to stop hardcoding RHEL8.
In this patch we can re-use an ansible fact to figure out which number
is actually deployed.

Also fixing ceph on OSP17.


mandre (commented):
/lgtm

mandre:
I wonder why ansible lint complains now. Nothing related to this patch though.

PR #193, opened 2022-06-17
network: remove 2-steps nmstate commit

EmilienM:
On CentOS9, NetworkManager CLI no longer reports the checkpoint name in
stdout so we can't rely on it anymore to do a 2 steps commit & apply.

Also using "apply" vs "set" since "set" is deprecated.


mdbooth:
You sure you want to do that? If it still does checkpoints there will we a way to use the feature.

The advantage of doing a checkpoint is that if you don't commit it within the timeout it automatically rolls it back. This saved me a *bunch* of redeploys in development when I was working on the networking stuff. I'd be inclined to look harder for how to keep it working with the new version.

EmilienM:
I agree about the advantage, but I haven't found out to get the checkpoint name in centos9, it's not in stdout anymore so we can't do like we do now and I haven't found any CLI which gives the name of the checkpoint...

EmilienM:
I finally found a way to make this work and keeping the checkpoint. This was tested on RHEL 9.2 but not on RHEL 8. I'll have to do it.

EmilienM:
tested on RHEL 8.4 and 9.2, it worked as expected.

PR #196, opened 2022-11-01
Allow to override MTUs

EmilienM:
* br-ex/public MTU default remains 1500, but is now overridable
* br-ctlplane/br-hostonly can be overridable and default is 1400
* Neutron MTU for tenant network remains 1300 by default but this needs
  to be bumped if double encapsulation is done with IPv6, see
  https://issues.redhat.com/browse/OCPBUGS-2921


MaysaMacedo:
the lint failure seems unrelated

mandre (approved):
LGTM. Lint failure unrelated as the patch hasn't changed anything in that area, but seems like something we'll need to look at.

PR #197, opened 2022-11-10
Fix MTU concatenation

MaysaMacedo:
This commit converts the mtu to int before concataneting, otherwise we would get the following error:

"Unexpected templating type error occurred on ({{ ctlplane_mtu + 100 }}): can only concatenate str (not \"int\") to str"}"


EmilienM:
Thanks Maysa, sorry totally my fault here.

PR #200, opened 2022-12-08
Increase default Neutron MTU

MaysaMacedo:
Due to a known issue with OVN-Kubernetes the default Neutron MTU should be at least 1400.


playbooks/vars/defaults.yaml
```diff
@@ -243,7 +243,7 @@ ovn_enabled: true
 # The default value can be problematic when doing double encapsulaiton with OVN,
 # the value will have to be increased. See the following bug report:
 # https://issues.redhat.com/browse/OCPBUGS-2921
-neutron_mtu: 1300
+neutron_mtu: 1400
```
EmilienM, line 246: please update the doc line 238

PR #201, opened 2023-03-13
prepare_host: add new RH CA certs

EmilienM:
We now have 2022 certs that we need to support alongside with the old
CA.


mandre (approved):
LGTM

PR #204, opened 2023-05-09
Allow to easily enable Barbican (with Simple Crypto)

EmilienM:
Allow to easily install Barbican with the simple crypto plugin.

This is opinionated but can easily be overridden.

NOTE: this PR also fixes a few lint errors.


mandre (commented):
/lgtm

PR #209, opened 2023-05-31
Updated support for rhsm builds of rhosp17 on rhel9

gprocunier:
Updated support for doing rhsm_enabled builds against official RHEL9/RHOS with registry.redhat.io
Added fixes for using service accounts via the terms based registry on access.redhat.com
Misc fixes for deploying virtually in libvirt/kvm
Provided an example local-overrides for rhos17 on rhel9 with details on creating the required components.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,9 @@ dev-install installs [TripleO standalone](https://docs.openstack.org/project-dep
 dev-install requires that:
 * an appropriate OS has already been installed
 * the machine running dev-install can SSH to the standalone host as either root or a user with passwordless sudo access
-* this machine has Ansible installed, and some dependencies like python3-netaddr.
+* this machine has Ansible installed, and some dependencies like python3-netaddr, python3-jmespath, python3-openstackclient.
 
-For OSP 16.2, the recommended OS is RHEL 8.4. For OSP 17, the recommended RHEL will be 9.
+For OSP 16.2, the recommended OS is RHEL 8.4. For OSP 17.0, the required RHEL will be 9.0.
 There is no need to do any other configuration prior to running dev-install.
 When deploying on TripleO from upstream, you need to deploy on CentOS Stream. If CentOS is not Stream, dev-install will migrate it.
 
--- a/playbooks/install_stack.yaml
+++ b/playbooks/install_stack.yaml
@@ -388,7 +388,10 @@
     #
     # Therefore it is easier to use the config.json method for podman.
     - name: Setup Docker registry authentication
-      when: cip_config is defined
+      when:
+        - cip_config is defined
+        - rhsm_enabled is defined
+        - registry_auth_file is defined
       become: true
       become_user: root
       block:
--- a/playbooks/install_stack.yaml
+++ b/playbooks/install_stack.yaml
@@ -389,9 +389,11 @@
     # Therefore it is easier to use the config.json method for podman.
     - name: Setup Docker registry authentication
       when: cip_config is defined
+      become: true
+      become_user: root
       block:
         - name: Ensure the .docker directory exists
-          file:
+          ansible.builtin.file:
             path: "/root/.docker"
             state: directory
             mode: '0755'
@@ -399,14 +401,12 @@
             group: "root"
 
         - name: Copy config.json to remote host
-          copy:
+          ansible.builtin.copy:
             src: "{{ registry_auth_file }}"
             dest: "/root/.docker/config.json"
             owner: "root"
             group: "root"
             mode: '0644'
-      become: true
-      become_user: root
 
     - name: Download containers-prepare-parameters.yaml from upstream master if cip_config is not defined on CentOS
       when:
--- a/playbooks/install_stack.yaml
+++ b/playbooks/install_stack.yaml
@@ -380,6 +380,33 @@
         cip_content:
           parameter_defaults:
             ContainerImagePrepare: "{{ cip_config }}"
+            ContainerImageRegistryLogin: true
+
+    # Red Hat Service Accounts have fixed usernames that incorporate a pipe character
+    # there is a lot of work required to handle this vs. a straight username because of to_nice_yaml
+    # when the rhcs5 portion of the build is started.
+    #
+    # Therefore it is easier to use the config.json method for podman.
+    - name: Setup Docker registry authentication
+      when: cip_config is defined
+      block:
+        - name: Ensure the .docker directory exists
+          file:
+            path: "/root/.docker"
+            state: directory
+            mode: '0755'
+            owner: "root"
+            group: "root"
+
+        - name: Copy config.json to remote host
+          copy:
+            src: "{{ registry_auth_file }}"
+            dest: "/root/.docker/config.json"
+            owner: "root"
+            group: "root"
+            mode: '0644'
+      become: true
+      become_user: root
 
     - name: Download containers-prepare-parameters.yaml from upstream master if cip_config is not defined on CentOS
       when:
--- a/playbooks/network.yaml
+++ b/playbooks/network.yaml
@@ -113,7 +113,7 @@
         stdin: "{{ network_state | to_nice_json }}"
       vars:
         network_state:
-          interfaces: "{{ nmstate_ifs }}"
+          interfaces: "{{ nmstate_ifs | from_yaml }}"
           # add saved static routes
           routes:
             config: "{{ nmstate_routes }}"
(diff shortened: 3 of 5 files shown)


playbooks/network.yaml
```diff
@@ -125,11 +125,11 @@
       ansible.builtin.set_fact:
         checkpoint: "{{ (nmstateset.stdout_lines | last).split()[1] }}"
 
-    - name: Set fact for nmstate checkpoing on RHEL9
+    - name: Set fact for nmstate checkpoint on RHEL9
       when:
         - ansible_facts.distribution_major_version == "9"
       ansible.builtin.set_fact:
-        checkpoint: "{{ (nmstateset.stderr_lines | last).split()[-1] }}"
+        checkpoint: "{{ (nmstateset.stdout_lines | select('search', 'Checkpoint: NetworkManager\\|') | last).split()[1] }}"
```
EmilienM, line 132: I've tested RHEL 9.2 and OSP 17.1 and this wasn't needed.

playbooks/prepare_host.yaml
```diff
@@ -37,14 +37,19 @@
...
         - name: Install container-tools module # noqa no-changed-when
           ansible.builtin.shell: |
             dnf module disable -y container-tools:rhel8
             dnf module enable -y container-tools:"{{ rhsm_container_tools_version }}"
+          when: rhsm_release | float < 9.0
         - name: Install virt module # noqa no-changed-when
           ansible.builtin.shell: |
             dnf module disable -y virt:rhel
             dnf module enable -y virt:"{{ virt_release }}"
+          when: rhsm_release | float < 9.0
```
EmilienM, line 52: I've tested RHEL 9.2 and OSP 17.1 and this wasn't needed.

playbooks/install_stack.yaml
```diff
@@ -380,6 +380,33 @@
         cip_content:
           parameter_defaults:
             ContainerImagePrepare: "{{ cip_config }}"
+            ContainerImageRegistryLogin: true
+
+    # Red Hat Service Accounts have fixed usernames that incorporate a pipe character
```
EmilienM, line 385: I'm not sure I understand why we need that now.

playbooks/prepare_host.yaml
```diff
@@ -37,14 +37,19 @@
         - name: Configure Red Hat Subscription Manager
           ansible.builtin.import_role:
             name: redhat-subscription
+        # this needs to be done early
+        - name: set rhsm_release
+          shell: subscription-manager release --set={{rhsm_release}}
```
EmilienM, line 42: Let's not do that. We have an Ansible role to manage RHSM, the release should be set by the role already with `rhsm_release`.

playbooks/prepare_host.yaml
```diff
@@ -37,14 +37,19 @@
...
         - name: Install container-tools module # noqa no-changed-when
           ansible.builtin.shell: |
             dnf module disable -y container-tools:rhel8
             dnf module enable -y container-tools:"{{ rhsm_container_tools_version }}"
+          when: rhsm_release | float < 9.0
```
EmilienM, line 47: we should use `ansible_facts.distribution_major_version`.
are you sure that this module won't exist on OSP 17.1?

playbooks/prepare_host.yaml
EmilienM, line 52: same here

playbooks/install_stack.yaml
gprocunier: regarding the ContainerImageRegistryLogin: true in the cip config, this is again because the rhcs5 piece is handled differently than the rest of the cloud install

this is again only for rhsm based installs

playbooks/network.yaml
```diff
@@ -113,7 +113,7 @@
         stdin: "{{ network_state | to_nice_json }}"
       vars:
         network_state:
-          interfaces: "{{ nmstate_ifs }}"
+          interfaces: "{{ nmstate_ifs | from_yaml }}"
```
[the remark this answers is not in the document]
gprocunier: Without that, this fails when deploying on kvm/libvirt.

playbooks/network.yaml
gprocunier: regarding the change to Set fact for nmstate checkpoint on RHEL9, this was required when i did a clean install of OSP 17.0 on RHEL9.0

it was not required on centos stream

it seems particular to 17.0 on RHEL9

playbooks/network.yaml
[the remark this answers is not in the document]
gprocunier: rhel 9.0 on OSP17.0

EmilienM:
rhel 9.0 on OSP17.0 is not something we'll support, we'll wait for 17.1 GA and see how things go.
The nmstate stuff will have to be tested again against 9.2 and 17.1, which is what I tested already and it worked fine. Although not in a VM.

playbooks/network.yaml
EmilienM: ack, will have to test it on RHEL 8.4 and RHEL 9.2 too to make sure it's compatible.

gprocunier: I have added handlers for rhel 9.0 + osp 17.0 and rhel 9.2 (and onwards) and osp 17.1

playbooks/prepare_host.yaml
gprocunier: official docs no longer have this step for rhos 17+ on rhel9+

playbooks/prepare_host.yaml
gprocunier: official docs no longer have this step for rhos 17+ on rhel9+

playbooks/install_stack.yaml
```diff
@@ -380,6 +380,33 @@
...
+          ansible.builtin.file:
+            path: "/root/.docker"
+            state: directory
+            mode: '0755'
+            owner: "root"
+            group: "root"
+
+        - name: Copy config.json to remote host
+          ansible.builtin.copy:
+            src: "{{ registry_auth_file }}"
```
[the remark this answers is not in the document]
gprocunier: Handled in the latest commit

README.md
```diff
@@ -7,9 +7,9 @@ dev-install installs [TripleO standalone](https://docs.openstack.org/project-dep
 dev-install requires that:
 * an appropriate OS has already been installed
 * the machine running dev-install can SSH to the standalone host as either root or a user with passwordless sudo access
-* this machine has Ansible installed, and some dependencies like python3-netaddr.
+* this machine has Ansible installed, and some dependencies like python3-netaddr, python3-jmespath, python3-openstackclient.
 
-For OSP 16.2, the recommended OS is RHEL 8.4. For OSP 17, the recommended RHEL will be 9.
+For OSP 16.2, the recommended OS is RHEL 8.4. For OSP 17.0, the required RHEL will be 9.0.
```
EmilienM, line 12: 17.1 being out soon, I would prefer that we document this one since it's the one that will be supported in long term.

Please change it to 17.1 and 9.2.

playbooks/network.yaml
EmilienM: If it's only a RHEL 9.0 issue, I suggest that we don't add this code, to reduce complexity.

I think dev-install is going to support 16.2 and OSP 17.1 in the longterm, we can't add support for every version of OSP, especially the unsupported ones (17.0).

playbooks/install_stack.yaml
```diff
@@ -380,6 +380,36 @@
...
           parameter_defaults:
             ContainerImagePrepare: "{{ cip_config }}"
+            ContainerImageRegistryLogin: true
+
+    # Red Hat Service Accounts have fixed usernames that incorporate a pipe character
+    # there is a lot of work required to handle this vs. a straight username because of to_nice_yaml
+    # when the rhcs5 portion of the build is started.
+    #
+    # Therefore it is easier to use the config.json method for podman.
+    - name: Setup Docker registry authentication
```
EmilienM, line 390: I'm currently investigating this too. Not saying I reject your solution here but I'm looking if we can do easier.

playbooks/prepare_host.yaml
```diff
@@ -41,10 +41,12 @@
           ansible.builtin.shell: |
             dnf module disable -y container-tools:rhel8
             dnf module enable -y container-tools:"{{ rhsm_container_tools_version }}"
+          when: ansible_distribution_major_version | int < 9
```
EmilienM, line 44: yes we need this change for sure.

playbooks/prepare_host.yaml
```diff
@@ -41,10 +41,12 @@
...
         - name: Install virt module # noqa no-changed-when
           ansible.builtin.shell: |
             dnf module disable -y virt:rhel
             dnf module enable -y virt:"{{ virt_release }}"
+          when: ansible_distribution_major_version | int < 9
```
EmilienM, line 49: here too we need this change.

EmilienM:
@gprocunier finally back to it (17.1 is now GA :partying_face:), I've proposed https://github.com/shiftstack/dev-install/pull/216

Let me know what you think, testing will follow soon when time permits.

EmilienM:
@gprocunier I'll close that PR if you don't mind, since we have PR#216 which address the OSP 17.1 use case.

PR #215, opened 2023-08-02
network: enable the dummy interfaces

EmilienM:
The dummy interfaces should be enabled if they have received a desired
route to a next hop.

This would avoid this kind of error:

```
NmstateError: InvalidArgument: The next hop interface of desired Route 'destination: 169.254.0.0/16 next-hop-interface: dummy0 next-hop-address: 0.0.0.0 metric: 1006 table-id: 254' has been marked as IPv4 disabled
```

### DIFF
--- a/playbooks/network.yaml
+++ b/playbooks/network.yaml
@@ -40,16 +40,16 @@
             type: dummy
             state: up
             ipv4:
-              enabled: false
+              enabled: true
             ipv6:
-              enabled: false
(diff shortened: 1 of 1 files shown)


vkmc:
Tested and got this output

"NmstateError: InvalidArgument: Interface dummy1 cannot have IP enabled as it is attached to a controller where IP is not allowed"

PR #216, opened 2023-08-23
:warning:  Switch OSP default to 17.1 :warning:

EmilienM:
* Update README and add RHEL versions table
* Move telco examples to 17.1
* Add new example for 16.2
* Updade vars defaults and add some doc too for storage on RHEL9
* Skip some tasks in prepare_host for RHEL9
* Add `ntp_server` parameter (with a safe default when not using RHSM) for deploying with a specific NTP.
* Handle the RH container registry login when deploying Ceph.
  Unfortunately, `redhat_registry_credentials` is now a dict with mandatory `username` and `password` fields.
* Make cephadm config working for DCN.
* Allow to expose `ceph_devices` via LVM.


EmilienM:
Some of us are currently testing it, I suggest that we wait for it to be merged.

Note that this brings backward incompatible changes to dev-install; so I'll produce a release today which will be usable for anyone who wants the default to work with 16.2 and RHEL 8.4.

EmilienM:
I need to add support for NtpServer when using internal puddles. That's for Ceph; I'll explain in a next patchset, tomorrow.

example-overrides/local-overrides-osp16-2-with-dpdk.yaml
```diff
@@ -1,9 +0,0 @@
-dpdk_interface: enp130s0f0
```
MaysaMacedo, line 1: why is this example being removed?

playbooks/vars/defaults.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-rhos_release: 16.2
+rhos_release: 17.1
```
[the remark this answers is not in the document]
EmilienM: yes I know, but at the same time I don't think we should wait for the next FDP release (October AFIK).
dev-install users who want FDP should know that it's currently not working now in 17.1...

example-overrides/local-overrides-osp16-2-with-dpdk.yaml
EmilienM: it's renamed for 17.1. No point of duplicating examples for DPDK.

MaysaMacedo:
@EmilienM Is it ready for review?

EmilienM:
> @EmilienM Is it ready for review?

Yes, I'm debugging something today but I'm not sure it involves dev-install.

EmilienM:
This is ready for review, @gprocunier had some comments (offline) but I think this can be addressed on a separate PR.

EmilienM:
further fixes/improvements should come via another PR.

Thanks for the review, let's iterate.

PR #223, opened 2024-01-24
Add ability to configure multiple DNS in DSAL helper

dulek:


mandre (approved):
I'm not using DSAL helper, but it may be useful to others.

PR #224, opened 2024-04-24
Checkout tripleo operators from latest commit

EmilienM:
The repo was retired, so we need to checkout the latest commit before it
gets empty.


mandre (commented):
/lgtm
/approve

mandre (approved):

PR #228, opened 2024-06-26
OSASINFRA-3495: Adding a StrageNFS Network to work with Manila/Ganesha and Ceph 

MiguelCarpio:
Adding a StrageNFS Network to work with Manila/Ganesha and Ceph

### DIFF
--- a/playbooks/prepare_stack.yaml
+++ b/playbooks/prepare_stack.yaml
@@ -153,6 +153,22 @@
         fi
       environment:
         OS_CLOUD: standalone
+      
+    - name: Create hostonly nfs network # noqa no-changed-when
+      when: manila_enabled
+      ansible.builtin.shell: |
+        if ! openstack network show hostonly-nfs; then
+            openstack network create --project openshift --share --external --provider-physical-network hostonly-nfs --provider-network-type flat hostonly-nfs
+        fi
+        if ! openstack subnet show hostonly-nfs-subnet; then
+            openstack subnet create --project openshift hostonly-nfs-subnet --subnet-range "{{ hostonly_nfs_cidr }}" \
+                --dhcp --gateway "{{ hostonly_nfs_gateway }}" \
+                --dns-nameserver "{{ network_info.dns | first }}" \
+                --allocation-pool "start={{ hostonly_nfs_fip_pool_start }},end={{ hostonly_nfs_fip_pool_end }}" \
+                --network hostonly-nfs
+        fi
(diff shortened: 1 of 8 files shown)


playbooks/prepare_stack.yaml
MiguelCarpio, line 1: Here, I'm creating the StorageNFS network using the default datacentre provider and the VLAN 70

playbooks/roles/ceph/tasks/storage-network.yml
MiguelCarpio, line 1: Here, I'm defining the StorageNFS network to be used by Ceph

playbooks/templates/standalone_parameters.ceph_ansible.yaml.j2
MiguelCarpio, line 1: Updating the standalone_parameters.ceph_ansible.yaml to use the hostonly_nfs_ganesha defined defaults.yaml

EmilienM:
from a very quick look, this sounds fine.
have you tested it?

MiguelCarpio:
> from a very quick look, this sounds fine. have you tested it?

Yes, I've tested it but I always get this error at `TASK | Render hieradata from template`:

### 2024-06-26 18:06:53.030809 | 78ac448c-2b24-54fe-e7f7-000000000490 |      FATAL | Render hieradata from template | dell-r640-100 | item=vip_data | error={"ansible_loop_var": "item", "changed": false, "item": "vip_data", "msg": "AnsibleUndefinedVariable: 'dict object' has no attribute 'storage_nfs'. 'dict object' has no attribute 'storage_nfs'"}

playbooks/templates/standalone_parameters.yaml.j2
```diff
@@ -46,6 +46,18 @@ resource_registry:
 {% endif %}
 parameter_defaults:
   CloudName: {{ hostname }}.{{ clouddomain }}
+  # Enable ganesha
```
EmilienM, line 49: this should be conditional whether manila is enabled

EmilienM:
I think during the tripleo deploy (after ceph itself), because you need to load a network file that has the StorageNFS definition. You have it for Ceph, but not for the tripleo deployment after.
Grep for `/usr/share/openstack-tripleo-heat-templates/network_data_undercloud.yaml`, we load that file when doing tripleo deploy. We could template this file and optionally add the StorageNFS if manila is enabled.

PR #233, opened 2025-03-06
Tweaks to client generation

stephenfin:
* Install necessary dependencies early in the process
* Install OSC with `dnf`, not `pip`
* Change cloud name generation so `standalone` becomes `standalone-admin` and `standalone_openshift` becomes `standalone`

I imagine the last one could be slightly contentious since it changes people's workflows, but it makes sense to have the most commonly used cloud have the shortest name.


EmilienM (approved):

EmilienM:
In your PR, standalone_openshift becomes openshift, right?

EmilienM:
why not calling is:
* `cloud-name`_admin for the admin project
* `cloud-name`_openshift for the openshift project

PR #235, opened 2025-11-19
Using registry credentials is now mandatory.

gryf:
From now on, to have working deployment, you'll need credentials for the registry with images for ceph and tripleo.


stephenfin (commented):
/approve
/lgtm

*EDIT:* nvm, no prow here :sweat_smile:

stephenfin (approved):

PR #238, opened 2026-01-26
Generate Python 3.13+ compatible certs

stephenfin:
Python 3.13 tightened SSL cert validation requirements [1][1] which requires changes to how we generate our CA and server CSRs [2][2].

Once merged, we will need to regenerate certs on existing hosts.

[1]: https://docs.python.org/3/whatsnew/3.13.html#ssl
[2]: https://galaxy.ansible.com/ui/repo/published/community/crypto/content/module/openssl_csr/


mandre (approved):
Pending test on an actual cloud.

mandre:
Tested it locally, works as expected.

PR #239, opened 2026-06-23
Fix broken conditionals: replace var|string with proper boolean checks

mandre:
Newer Ansible versions (2.17+) require 'when' conditionals to evaluate to a boolean, not a truthy string. The |string filter converts a value to its string representation but does not produce a boolean result.

Replace all 'var|string' conditionals with
'var | default('', true) | length > 0' (positive checks) and 'var | default('', true) | length == 0' (negated checks) to produce proper boolean results.


gryf (approved):
/lgtm

PR #240, opened 2026-06-23
Fix container registry credentials not passed when rhsm is disabled

mandre:
ContainerImageRegistryCredentials and ContainerImageRegistryLogin were gated behind {% if rhsm_enabled %}, so when using the rhos-release path they were never rendered into standalone_parameters.yaml. This caused TripleO to fail pulling container images with an authentication error, even though podman login succeeded separately.

Decouple the credentials block from rhsm_enabled and gate it on registers being defined instead, so credentials are passed to TripleO whenever registries are configured regardless of subscription method.


gryf (commented):
/lgtm

gryf (approved):

PR #241, opened 2026-06-23
Fix reboot task to escalate as root instead of stack

mandre:
The reboot task had become: true but inherited become_user: stack from the play level, causing 'Interactive authentication required' from polkit/systemd. Add become_user: root to match the pattern used by all other root-privileged tasks in the playbook.


gryf (approved):
Looks good. We can ignore linting error for now, as `redhat-subscription` should be optional.

PR #246, opened 2026-08-26
Add configuration flexibility for network and MTU settings

chenwng:
This PR fixes following issues I found during deployment.

* os-net-config fails when the max_mtu of dummy interfaces is 0 .  Add `enable_dummy_mtu` to skip setting mtu for dummy interfaces.
* If `external_fip_pool_start` and `external_fip_pool_end` are not defined, network `external` will not be created. Running `make prepare_stack_testconfig` will fail when setting the gateway for the private router.  Add `external_network` to allow using `hostonly`.
* If `rhsm_enabled` is not true, `NtpServer` in `playbooks/templates/standalone_parameters.yaml.j2` will not be set, which results chrony sync failure during deployment in internal network. Move `NtpServer` out of `rhsm_enabled` check to allow setting internal ntp server.

### DIFF
--- a/.github/workflows/ansible-lint.yml
+++ b/.github/workflows/ansible-lint.yml
@@ -19,5 +19,10 @@ jobs:
       - name: Install dependency
         run: ansible-galaxy collection install -r requirements.yaml
 
+      # This works around the lint error from `ansible-playbook --syntax-check -vv playbooks/install_stack.yaml`
+      # [ERROR]: Could not find specified file in role, its 'tasks/' is not usable.
+      - name: Prepare redhat-subscription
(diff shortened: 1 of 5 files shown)


mandre:
I believe I've already fixed this issue last time I deployed with dev-install. I'm not in front of the computer right now so can't confirm, have a look at the open PRs in this repo.

chenwng:
I found #245  tries to fix the mtu issue.

gryf:
I would split this PR into multiple ones addressing single issue per PR.

chenwng:
Closing as most of them have been fixed by PRs from Martin.

